PR #11, opened 2014-12-16
OCLM-10: Implement tests for retiring concepts and modifying description...

jbuczynski:
...s


rkorytkowski:
It's not the most elegant, but good enough :) Thanks a lot! Sorry for the delay.

PR #13, opened 2015-01-08
Limit the number of rows to 100 in the detail page OCLM-20

ningosi:


api/src/main/java/org/openmrs/module/openconceptlab/UpdateService.java
```diff
@@ -258,4 +261,32 @@ public void scheduleUpdate() {
...
+
+	/**
+	 *
+	 * @param update the update to be passed
+	 * @param first starting index
+	 * @param max  maximum limit
+	 * @return a list of items
+	 */
+	public List<Item> getUpdateItems(Update update, int first, int max) {
+		List<Item> list = new ArrayList<Item>(update.getItems());
```
rkorytkowski: This must be implemented using Hibernate Criteria to be efficient. We cannot afford to read 40k+ items from the db to return just a small subset of them.

ningosi: @rkorytkowski if you have time plaese help me understand how to link that criteria with already existing Update object

PR #14, opened 2015-03-03
OCLM-21

ningosi:
@rkorytkowski please review this

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/UpdateService.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/UpdateService.java
@@ -19,6 +19,7 @@ import org.hibernate.Criteria;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.criterion.Order;
+import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Restrictions;
 import org.openmrs.GlobalProperty;
 import org.openmrs.api.AdministrationService;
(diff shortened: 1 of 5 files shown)


omod/src/main/java/org/openmrs/module/openconceptlab/fragment/controller/status/PreviousUpdatesFragmentController.java
```diff
@@ -45,27 +43,23 @@ public void controller(FragmentModel model,
...
-								duration = Utils.dateDifference(update.getLocalDateStarted(), update.getLocalDateStopped(), TimeUnit.MINUTES).intValue();
-							}
-						}
-						String status;
+						Set<State> states = new HashSet<State>();
+						states.add(State.ERROR);
+
+						Integer errors = service.getUpdateItemsCount(update, states);
+						Integer totalItems =  service.getUpdateItemsCount(update, new HashSet<State>());
+						//loop through each item object to count error
```
rkorytkowski: This comment is no longer true :)

api/src/main/java/org/openmrs/module/openconceptlab/UpdateService.java
```diff
@@ -284,4 +286,19 @@ public void scheduleUpdate() {
...
+	 * @return a count of items
+	 */
+	public Integer getUpdateItemsCount(Update update, Set<State> states) {
+		Criteria items = getSession().createCriteria(Item.class);
+		items.add(Restrictions.eq("update", update));
+		items.addOrder(Order.desc("state"));
+		if (!(states.isEmpty())) {
+			items.add(Restrictions.in("state", states));
+		}
+		return  ((Long) items.setProjection(Projections.rowCount()).uniqueResult()).intValue();
```
rkorytkowski: It's safer to cast it to Number instead of Long. In some older OpenMRS versions it is Integer not Long.

PR #16, opened 2015-07-07
Removed failing changeset

ningosi:
@rkorytkowski  removed the failing change set

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
@@ -20,7 +20,11 @@ import org.apache.commons.httpclient.util.DateParseException;
 import org.apache.commons.httpclient.util.DateUtil;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.openmrs.module.openconceptlab.Update;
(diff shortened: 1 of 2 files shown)


dkayiwa:
@rkorytkowski do you think this pull request is ready for merging? :)

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -67,12 +75,19 @@ public OclResponse fetchUpdates(String url, String token, Date updatedSince) thr
...
+
+		//create a while loop to check if the last update contained items more than 1000
+		Integer pages = 1;
+		Update lastUpdate = updateService.getLastUpdate();
+		while(lastUpdate.getItems().size() >= 1000) {
+			pages++;
+			get.getParams().setParameter("page", pages);
+			client.executeMethod(get);
+		}
+
```
rkorytkowski: It won't work. You do fetch all pages, but do not process all results... The results are processed only from the last fetched page in extractResponse(get). You should rather call a new cacheResponse method inside the while loop and make it write all pages to the same zipped file on disk and then return OclResponse for that file.

Also at this point you should not check getItems().size(), because it will get updated only after items are processed (installed). Instead you could simply get the next page and break the loop if the results do not contain any concepts.

PR #24, opened 2016-10-28
OCLM-38 - Subscribe to snapshot or released versions

tmarzeion:
What is done here:
- column `release_version` added to `openconceptlab_update` table, it is used to check if latest OCL release is newer than one downloaded by OpenMRS OCL Client.
- `fetchUpdates()` method in `OclClient` is now renamed to `fetchSnapshotUpdates()`.
- `fetchSnapshotUpdates()` is called only when `isSubscribedToSnapshot()` in `Subscription` returns `true`.
- `fetchInitialUpdates()` is renamed to `fetchReleaseUpdates()`
- `fetchReleaseUpdates()` overloaded method is added, this method checks if last downloaded version is up to date with OCL release version. If yes, old `fetchReleaseUpdates()` is called.

I've added 3 test methods used to start update.
`startUpdate_shouldStartInitialUpdate()` - Starts initial update
`startUpdate_shouldStartReleaseUpdate()` - Starts followup update (only release version updates)
`startUpdate_shouldStartSnapshotUpdate()` - Starts followup update (with SNAPSHOT versions included)
These methods are `@Ignored` because tests are fetching data from original source, so it takes a lot of time (about 13 mins per test) to finish these tests. This issue will be fixed when some custom, small sized test-exclusive data will be created.

There is also `update_shouldUpdateLatestDownloadedRelease()` test method, which checks if new `String lastDownloadedRelease` field in `Update` object is successfully saved into DB.

**Summary:**
This code works like it worked before, until `boolean isFetchingSnapshotUpdates` in `Subscription` isn't changed to `false`, which is going to be implemented in new UI.
when `isFetchingSnapshotUpdates` is set to false, there is new logic triggered in `Updater`, which calls `OclClient` overloaded `fetchReleaseUpdates()` method that does update if new source version is released.

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/Import.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/Import.java
@@ -45,6 +45,10 @@ public class Import {
 	@Basic
 	@Column(name = "ocl_date_started")
 	private Date oclDateStarted;
+
+	@Basic
+	@Column(name = "release_version")
+	private String releaseVersion;
 	
 	@Basic
 	@Column(name = "error_message")
@@ -80,6 +84,14 @@ public class Import {
 	
 	public void setOclDateStarted(Date oclDateStarted) {
 		this.oclDateStarted = oclDateStarted;
+	}
+
+	public String getReleaseVersion() {
+		return releaseVersion;
+	}
+
+	public void setReleaseVersion(String releaseVersion) {
+		this.releaseVersion = releaseVersion;
 	}
 	
     public String getErrorMessage() {
--- a/api/src/main/java/org/openmrs/module/openconceptlab/ImportService.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/ImportService.java
@@ -59,6 +59,8 @@ public interface ImportService {
     @Transactional
 	void updateOclDateStarted(Import update, Date oclDateStarted);
 
+	void updateReleaseVersion(Import anImport, String version);
+
 	/**
 	 * @should throw IllegalArgumentException if not scheduled
 	 * @should throw IllegalStateException if trying to stop twice
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
@@ -9,6 +9,18 @@
  */
 package org.openmrs.module.openconceptlab.client;
 
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.httpclient.Header;
+import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.HttpException;
+import org.apache.commons.httpclient.NameValuePair;
+import org.apache.commons.httpclient.methods.GetMethod;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.openmrs.util.OpenmrsUtil;
+
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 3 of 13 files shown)


tmarzeion:
@rkorytkowski 
Code is updated, 
I've added 3 test methods used to start update.
`startUpdate_shouldStartInitialUpdate()` - Starts initial update
`startUpdate_shouldStartReleaseUpdate()` - Starts followup update (only release version updates)
`startUpdate_shouldStartSnapshotUpdate()` - Starts followup update (with SNAPSHOT versions included)
These methods are `@Ignored` because tests are fetching data from original source, so it takes a lot of time (about 13 mins per test) to finish these tests. This issue will be fixed when some custom, small sized test-exclusive data will be created.

There is also `update_shouldUpdateLatestDownloadedRelease()` test method, which checks if new `String lastDownloadedRelease' field in `Update` object is successfully saved into DB.

**Summary:**
This code works like it worked before, until `boolean isFetchingSnapshotUpdates` in `Subscription` isn't changed to `false`, which is going to be implemented in new UI.
when `isFetchingSnapshotUpdates` is set to false, there is new logic triggered in `Updater`, which calls `OclClient` overloaded `fetchReleaseUpdates()` method that does update if new source version is released.


# TL;DR - SHOULD WORK

rkorytkowski:
It's best you reapply the changes on top of what @AdamGrzybkowski did and address my review comment.

api/src/main/java/org/openmrs/module/openconceptlab/Import.java
```diff
@@ -45,6 +45,10 @@
 	@Basic
 	@Column(name = "ocl_date_started")
 	private Date oclDateStarted;
+
+	@Basic
+	@Column(name = "last_downloaded_release")
+	private String lastDownloadedRelease;
```
rkorytkowski: I asked to rename the variable to releaseVersion. As I understand it's not only downloaded, but release version imported for this import.

api/src/test/java/org/openmrs/module/openconceptlab/UpdateServiceTest.java
```diff
@@ -136,12 +146,108 @@ public void saveSubscription_shouldSaveSubscription() throws Exception {
...
+
+		importService.saveSubscription(newSubscription);
+        Importer importer = new Importer();
+
+        File tempDir = File.createTempFile("ocl", "");
+        FileUtils.deleteQuietly(tempDir);
+        tempDir.deleteOnExit();
+        OclClient oclClient = new OclClient(tempDir.getAbsolutePath());
+
+        importer.setOclClient(oclClient);
```
rkorytkowski: In addition to those tests let's actually have a test, which mocks OclClient.

api/src/main/java/org/openmrs/module/openconceptlab/Import.java
tmarzeion: Renamed

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,7 +123,17 @@ public OclResponse fetchInitialUpdates(String url, String token) throws IOExcept
 		
 		return extractResponse(exportUrlGet);
     }
-	
+
+	public OclResponse fetchReleaseUpdates(String url, String token, String lastDownloadedReleaseVersion) throws IOException {
```
rkorytkowski: lastDownloadedReleaseVersion -> lastReleaseVersion

api/src/main/java/org/openmrs/module/openconceptlab/ImportService.java
```diff
@@ -59,6 +59,8 @@
     @Transactional
 	void updateOclDateStarted(Import update, Date oclDateStarted);
 
+	void updateLastRelease(Import anImport, String version);
```
rkorytkowski: Didn't I say to change the name to updateReleaseVersion?

rkorytkowski:
@tmarzeion, are you having hard time following new github's review?

tmarzeion:
@rkorytkowski There is something wrong with this review, since I didn't see any comments before (for example, first review was requested yesterday at 11:22, but comments for this review were created today about 11:45)

PR #25, opened 2016-11-02
OCLM-37 OWA, REST - WIP

AdamGrzybkowski:
@rkorytkowski This is not finished yet, but if you have some time, please take a look and give me some feedback :)


omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    protected PageableResult doGetAll(RequestContext context) throws ResponseException {
+        return new NeedsPaging<Import>(getImportService().getImportsInOrder(0, 20), context);
+    }
+
+    @PropertyGetter("importProgress")
```
AdamGrzybkowski: @rkorytkowski Could you please take a took a this? I'm getting error when requesting full representation. [here](http://pastebin.com/Vs1EFKKe) is the stacktrace.

rkorytkowski:
Please merge as soon as possible :)

api/src/main/resources/liquibase.xml
```diff
@@ -143,5 +143,56 @@
...
+			<column name="error_message" type="varchar(1024)" />
+		</createTable>
+	</changeSet>
+
+	<changeSet id="openconceptlab-v1-02.11.2016-11.52" author="adamgrzybkowski">
+		<preConditions>
+			<tableExists tableName="openconceptlab_import"/>
+		</preConditions>
+		<comment>Add foreign key</comment>
+		<addForeignKeyConstraint constraintName="openconceptlab_item_import" baseTableName="openconceptlab_item" baseColumnNames="import_id" referencedTableName="openconceptlab_import" referencedColumnNames="import_id"/>
```
rkorytkowski: It will fail if there were any previous imports in the db. Imported items would reference non existing imports. I would drop both tables and simply recreate them. The module wasn't used in production anyway.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
rkorytkowski: Try changing returned value to Integer and argument to Import instead of Subscription.

AdamGrzybkowski: The problem was with argument. Thanks :)

api/src/main/resources/liquibase.xml
AdamGrzybkowski: Done

AdamGrzybkowski:
@rkorytkowski I've update PR. 
I have done most of the UI I guess, but.. 
Starting import via REST is not working, so I wasn't able to test the UI and the item resource. It would be great if you could take a look at ImportResource.save() metchod. I know this ticket is a blocker so I'm doing my best to end it ASAP. There still no test for the REST and OWA.

rkorytkowski:
I decided to merge in the code to unblock work on other issues.

rkorytkowski:
Let's address the issue with starting an update in a follow up PR. Could you please share what stacktrace you get?

PR #26, opened 2016-11-04
OCLM-37 follow up commit - WIP

AdamGrzybkowski:
@rkorytkowski I'm opening as a follow up commit. This is not finished yet. 
About the starting import problem, [here](http://pastebin.com/iZ8Vbtaf) is the stacktrace.


rkorytkowski:
Did you set subscription url to http://openconceptlab.com/orgs/CIEL/sources/CIEL/ ?

AdamGrzybkowski:
@rkorytkowski yes

rkorytkowski:
well I guess you need to try debugging at org.openmrs.module.openconceptlab.client.OclClient.fetchLatestVersion(OclClient.java:328) and see if the url is picked up correctly

AdamGrzybkowski:
Ok, I didn't set the correct url... but the one you provided isn't correct also :D 
The correct one that works is  http://api.openconceptlab.com/orgs/CIEL/sources/CIEL/

rkorytkowski:
That makes me think we need another issue to prepend "api." if it's not
there :)

PR #36, opened 2016-11-09
OCLM-47 Introduce import without validation 

AdamGrzybkowski:
@rkorytkowski Take a look at this [stacktrace](http://pastebin.com/BESFH7yn) when importing concepts without validation. Is this how you thought it's gonna work ?


AdamGrzybkowski:
@rkorytkowski I'm not really sure why but when I used method from importService to save concept it works without any errors. Before that I was injecting sessionFactory to the Saver and this was causing NPE.

api/src/test/java/org/openmrs/module/openconceptlab/importer/ImporterTest.java
```diff
@@ -9,1006 +9,217 @@
...
-			for (T element : target) {
-				if (!predicate.apply(element))
-					continue;
-				result = element;
-				break;
-			}
-			return result;
-		}
-
-	}
```
rkorytkowski: What happened to this file? Did you just accidentally changed formatting?

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -318,6 +318,9 @@ public Subscription getSubscription() {
 		String token = adminService.getGlobalProperty(OpenConceptLabConstants.GP_TOKEN);
 		subscription.setToken(token);
 
+		String validation = adminService.getGlobalProperty(OpenConceptLabConstants.GP_VALIDATION);
+		subscription.setValidation(Boolean.valueOf(validation));
```
rkorytkowski: It shouldn't be boolean as we will want to introduce more validation options. Make it an enum with "NONE" and "FULL" for now.

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
```diff
@@ -28,10 +28,18 @@
...
 
-	private boolean subscribedToSnapshot = true;
+	private Boolean subscribedToSnapshot = false;
 
-	public Subscription() {
+	private Boolean validation = true;
 
+	public Subscription() {}
+
+	public Boolean isValidation() {
```
rkorytkowski: "is" needs to return boolean and not Boolean for POJOs. Anyway it should be an enum type.

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
```diff
@@ -86,11 +94,11 @@ public boolean isManual() {
 		return (days == null && hours == null && minutes == null) || (days == 0 && hours == 0 && minutes == 0);
 	}
 
-	public boolean isSubscribedToSnapshot() {
+	public Boolean isSubscribedToSnapshot() {
```
rkorytkowski: "is" -> always returns boolean and not Boolean, "get" -> returns Boolean (it's the POJO contract), please revert it back.

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
AdamGrzybkowski: Yeah I did the enum first but then i noticed that you wanted a checkbox(validate or not) so I changed it to boolean.

api/src/test/java/org/openmrs/module/openconceptlab/importer/ImporterTest.java
AdamGrzybkowski: I have renamed tests classes to match renamed classes in some earlier PR.

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
AdamGrzybkowski: When I change it to boolean I'm getting some validation exception, I am not able to post Subscription at all because of this.

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
AdamGrzybkowski: @rkorytkowski done

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
```diff
@@ -86,11 +94,11 @@ public boolean isManual() {
 		return (days == null && hours == null && minutes == null) || (days == 0 && hours == 0 && minutes == 0);
 	}
 
-	public boolean isSubscribedToSnapshot() {
+	public Boolean getSubscribedToSnapshot() {
```
rkorytkowski: Didn't I say to use "boolean isSubscribedToSnapshot()"? It's better as you don't have to check for null, ever! :)

rkorytkowski:
Please apply the small change I requested and merge unless there are other issues I missed and you are aware of :)

api/src/main/java/org/openmrs/module/openconceptlab/Subscription.java
AdamGrzybkowski: Yes you did and I agree with it. But like i responded to you, I'm getting some validation error in webservices abstraction classes when using boolean

AdamGrzybkowski:
@rkorytkowski I've managed to use boolean instead of Boolean, but I had to use PropertySetter. I will wait with the merge for your comment on that :)

PR #37, opened 2016-11-09
OCLM-39/OCLM-49 - Support offline updates

tmarzeion:
### UI:
- [x] Add file update option
- [x] Show notifications when something happen
- [x] Show update progress

**Notes:**
- UI uses `"ng-file-upload": "^12.2.13"` to load file uploading component.
See https://github.com/danialfarid/ng-file-upload for details
### Backend:
- [x] File uploading via REST
- [x] Pass uploaded file to Importer
- [x] Import concepts/mappings
- [x] ~~Include a source version in deployment artifacts and load it at system startup~~ - moved to [OCLM-49](https://issues.openmrs.org/browse/OCLM-49)

**Notes:**
- File uploading by REST is done by passing file param to ImportResource, which now implements Uploadable interface.

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
@@ -10,13 +10,21 @@
 package org.openmrs.module.openconceptlab;
 
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.openmrs.GlobalProperty;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.DaemonToken;
 import org.openmrs.module.DaemonTokenAware;
 import org.openmrs.module.ModuleActivator;
+import org.openmrs.module.openconceptlab.importer.Importer;
 import org.openmrs.module.openconceptlab.scheduler.UpdateScheduler;
+import org.openmrs.util.OpenmrsUtil;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.zip.ZipFile;
 
 /**
  * This class contains the logic that is run every time this module is either started or stopped.
@@ -41,10 +49,46 @@ public class OpenConceptLabActivator implements ModuleActivator, DaemonTokenAwar
 		if (!Context.isSessionOpen()) {
 			Context.openSession();
 		}
-			
+
+		String loadAtStartupPath = Context.getAdministrationService()
+				.getGlobalProperty(OpenConceptLabConstants.GP_OCL_LOAD_AT_STARTUP_PATH);
+		if (StringUtils.isBlank(loadAtStartupPath)) {
+			loadAtStartupPath =
+					new File(new File(new File(OpenmrsUtil.getApplicationDataDirectory(), "ocl"), "configuration"), "loadAtStartup").getAbsolutePath();
+			Context.getAdministrationService()
+					.saveGlobalProperty(
+							new GlobalProperty(OpenConceptLabConstants.GP_OCL_LOAD_AT_STARTUP_PATH, loadAtStartupPath)
+					);
+		}
+
+		File loadAtStartupDir = new File(loadAtStartupPath);
+		if (!loadAtStartupDir.exists()) {
+			loadAtStartupDir.mkdirs();
+		}
+
 		UpdateScheduler scheduler = Context.getRegisteredComponent("openconceptlab.updateScheduler", UpdateScheduler.class);
-		scheduler .scheduleUpdate();
-		
+
+		File[] files = loadAtStartupDir.listFiles();
+		if (files != null && files.length > 1) {
+			throw new IllegalStateException(
+					"There is more than one file in ocl/loadAtStartup directory\n" +
+					"Ensure that there is only one file\n" +
+					"Absolute directory path: " + loadAtStartupDir.getAbsolutePath());
+		} else if (files != null && files.length != 0) {
+			if (files[0].getName().endsWith(".zip")) {
+				try {
+					ZipFile zipFile = (new ZipFile(files[0]));
+					Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+					importer.setImportFile(zipFile);
+					scheduler.scheduleNow();
+				} catch (IOException e) {
+					throw new IllegalStateException("Failed to open zip file", e);
+				}
+			} else {
+				throw new IllegalStateException("File " + files[0].getName() + " must be in *.zip format");
+			}
+		}
+		scheduler.scheduleUpdate();
 		log.info("Open Concept Lab Module refreshed");
 	}
 	
--- a/api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
@@ -35,6 +35,7 @@ import org.openmrs.module.openconceptlab.client.OclConcept;
 import org.openmrs.module.openconceptlab.client.OclMapping;
 import org.openmrs.module.openconceptlab.scheduler.UpdateScheduler;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
@@ -75,6 +76,8 @@ public class Importer implements Runnable {
 
 	private volatile long totalBytesToProcess;
 
+	private ZipFile importFile;
+
 	private interface Task {
 		void run() throws Exception;
 	}
@@ -95,6 +98,10 @@ public class Importer implements Runnable {
 	    this.saver = persister;
     }
 
+	public void setImportFile(ZipFile importFile) {
+		this.importFile = importFile;
+	}
+
 	/**
 	 * Runs an import for a configured subscription.
 	 * <p>
@@ -107,16 +114,16 @@ public class Importer implements Runnable {
 	 */
 	@Override
 	public void run() {
-		if (in == null) {
+		if (importFile != null) {
+			run(importFile);
+		}
+		else {
 			Daemon.runInDaemonThreadAndWait(new Runnable() {
 				@Override
 				public void run() {
 					runTask();
 				}
 			}, OpenConceptLabActivator.getDaemonToken());
-		}
-		else {
-			run(in);
 		}
 	}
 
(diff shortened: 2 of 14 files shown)


rkorytkowski:
The PR in webservices.rest is merged in https://github.com/openmrs/openmrs-module-webservices.rest/commit/cc4e47efb33aa6811ec42fbd931f571a3cf87553

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
```diff
@@ -173,4 +178,27 @@ private UpdateScheduler getUpdateScheduler() {
...
         return Context.getService(ImportService.class);
     }
+
+    @Override
+    public Object upload(MultipartFile multipartFile, RequestContext requestContext) throws ResponseException, IOException {
+
+        ImportService importService = getImportService();
+
+        Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+        importer.setImportService(importService);
```
rkorytkowski: You shouldn't need to set importService here. It's done by Spring for you.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
```diff
@@ -173,4 +178,27 @@ private UpdateScheduler getUpdateScheduler() {
...
+
+        File file = new File(multipartFile.getOriginalFilename());
+        multipartFile.transferTo(file);
+        ZipFile zipFile = new ZipFile(file);
+
+        importer.run(zipFile);
```
rkorytkowski: We shouldn't be importing anything synchronously. See UpdateScheduler.scheduleNow(). You need to find a way to pass the file to be imported along to the scheduler for the importer to process it.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
```diff
@@ -173,4 +178,27 @@ private UpdateScheduler getUpdateScheduler() {
...
+    @Override
+    public Object upload(MultipartFile multipartFile, RequestContext requestContext) throws ResponseException, IOException {
+
+        ImportService importService = getImportService();
+
+        Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+        importer.setImportService(importService);
+
+        File file = new File(multipartFile.getOriginalFilename());
+        multipartFile.transferTo(file);
```
rkorytkowski: Did you think of what happens to a file once it is processed? We don't want to keep them forever. It would be easier not to save the file to disk, rather keep it in memory. No need to worry about discarding it that way. The uploaded file should easily fit in memory CIEL ~50MB.

tmarzeion: @rkorytkowski 
Done :)

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
tmarzeion: @rkorytkowski 
Done :)

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
tmarzeion: Done :)
I did this by setting `InputStream` from export file to `InputStream` field of `Importer` class and calling appropriate method by using some condition logic.

rkorytkowski: @tmarzeion, are you working on this? My suggestion would be to actually store the file in memory in ByteArrayInputStream (MultipartFile is discarded after request is compete). You could add updateScheduler.scheduleNow(InputStream inputStream), which would create a new Import object and add a reference to point to that stream on the import object. Configure importer to use that Import object by adding Importer.setImport(Import anImport) (note you will have to adjust runAndHandleErrors to use that object when calling startImport).

dkayiwa:
@tmarzeion do you mind squashing these commits into one?

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -41,10 +50,51 @@ public void contextRefreshed() {
...
+					Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+					importer.setInputStream(inputStream);
+					scheduler.scheduleNow();
+					file.delete();
+				} catch (IOException e) {
+					log.error("Failed to open zip file", e);
+				}
+			} else {
+				log.error("File " + file.getName() + " must be in *.zip format");
+			}
```
rkorytkowski: Instead of logging errors, please throw exceptions. They should prevent the module from starting up and the error will be better communicated to the sys admin.

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -41,10 +50,51 @@ public void contextRefreshed() {
...
+		} else {
+			scheduler.scheduleUpdate();
```
rkorytkowski: I feel like this line should be called regardless... move it out of else.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
```diff
@@ -171,13 +179,43 @@ public void run() {
...
-						in = new CountingInputStream(inputStream);
-
-						importService.updateSubscriptionUrl(anImport, IMPORTED_VIA_API);
+						if (in != null) {
+							if (!in.equals(inputStream)) {
+								setInputStream(inputStream);
+							}
+							importService.updateSubscriptionUrl(anImport, IMPORTED_VIA_API);
+							//TODO Is this date OK?
+							Date date = new Date();
```
rkorytkowski: Good question. I guess we need to stay with that unless we can find the date inside the file...

tmarzeion:
@rkorytkowski Changes applied according to Your review :)

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
tmarzeion: Throwing IllegalStateException doesn't prevent module from starting up... What am I doing wrong? :crying_cat_face:

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -41,10 +50,48 @@ public void contextRefreshed() {
 		if (!Context.isSessionOpen()) {
 			Context.openSession();
 		}
-			
+
+		String loadAtStartupPath = Context.getAdministrationService()
+				.getGlobalProperty(OpenConceptLabConstants.GP_OCL_LOAD_AT_STARTUP_PATH);
+		if (StringUtils.isBlank(loadAtStartupPath)) {
+			loadAtStartupPath = new File(new File(OpenmrsUtil.getApplicationDataDirectory(), "ocl"), "loadAtStartup").getAbsolutePath();
```
tmarzeion: @dkayiwa 
Oops :fearful: 
Fixed :)

dkayiwa:
@rkorytkowski do you think this is now ready to merge? :)

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -41,10 +50,49 @@ public void contextRefreshed() {
...
+		} else if (files != null && files.length != 0) {
+			File file = files[0];
+			if (file.getName().endsWith(".zip")) {
+				InputStream inputStream;
+				try {
+					inputStream = Utils.extractExportInputStreamFromZip(new ZipFile(file));
+					Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+					importer.setInputStream(inputStream);
+					scheduler.scheduleNow();
+					file.delete();
```
rkorytkowski: You don't copy the inputStream... yet you delete the file... How is the importer going to read the file then?

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
```diff
@@ -171,13 +179,42 @@ public void run() {
 
 					@Override
 					public void run() throws IOException {
-						in = new CountingInputStream(inputStream);
-
-						importService.updateSubscriptionUrl(anImport, IMPORTED_VIA_API);
+						if (in != null) {
+							if (!in.equals(inputStream)) {
+								setInputStream(inputStream);
+							}
```
rkorytkowski: I don't quite follow what happens here... What's the purpose of this 'if'?

api/src/test/java/org/openmrs/module/openconceptlab/ImportServiceTest.java
```diff
@@ -311,7 +311,9 @@ public void run_shouldSetSubscriptionUrl() {
 		importer.setSaver(saver);
 
 		TestResources.setupDaemonToken();
-		importer.run(TestResources.getSimpleResponseAsJsonStream());
+		InputStream in = TestResources.getSimpleResponseAsJsonStream();
+		importer.setInputStream(in);
+		importer.run(in);
```
rkorytkowski: Why you first call setInputStream and then 'run' with the in parameter again? It seems like the API is not right.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
```diff
@@ -173,4 +180,25 @@ private UpdateScheduler getUpdateScheduler() {
...
     }
+
+    @Override
+    public Object upload(MultipartFile multipartFile, RequestContext requestContext) throws ResponseException, IOException {
+
+        File file = new File(multipartFile.getOriginalFilename());
+        multipartFile.transferTo(file);
+        ZipFile zipFile = new ZipFile(file);
+        InputStream inputStream = Utils.extractExportInputStreamFromZip(zipFile);
+        file.delete();
```
rkorytkowski: Same issue here. You tell to delete the file yet you pass the inputStream of the deleted file to the importer. It doesn't seem right. It feels like the importer should handle deletion. To achieve that instead of passing inputStream you would have to pass the file, process it and delete at the end.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
tmarzeion: @rkorytkowski I don't remember what was purpose of this, I've removed these conditions and it works fine anyway.

api/src/test/java/org/openmrs/module/openconceptlab/ImportServiceTest.java
tmarzeion: @rkorytkowski This one was caused by useless conditionals that you've mentioned before. Removed.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/resources/ImportResource.java
rkorytkowski: Did you verify the file is deleted? Most likely you open up a stream and OS blocks the file for read/write until you close it. The delete operation will not take effect.

tmarzeion: File gets deleted

PR #53, opened 2017-02-17
OCL-59 Use reflection to set isAllowDecimal on ConceptNumeric

AdamGrzybkowski:


api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -227,6 +233,19 @@ public Concept toConcept(CacheService cacheService, OclConcept oclConcept) throw
 		return concept;
 	}
 
+	private void setAllowDecimalUsingReflection(ConceptNumeric numeric, Extras extras) {
+		try {
+            Method setAllowDecimal = numeric.getClass().getDeclaredMethod("setAllowDecimal", Boolean.class);
+            setAllowDecimal.invoke(numeric, extras.getPrecise());
```
rkorytkowski: How about you call BeanUtils.setProperty() instead?

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -205,7 +207,11 @@ public Concept toConcept(CacheService cacheService, OclConcept oclConcept) throw
 
 			numeric.setUnits(extras.getUnits());
 
-			numeric.setPrecise(extras.getPrecise());
+			try {
+				numeric.setPrecise(extras.getPrecise());
+			} catch (NoSuchMethodError e) {
+				setAllowDecimalUsingReflection(numeric, extras);
```
rkorytkowski: How about you call BeanUtils.setProperty() instead?

PR #55, opened 2019-07-26
OCLM-67: Add support for importing names of type Index Term and Short

Karuhanga:
### JIRA TICKET NAME:
[OCLM-67: Add support for importing names of type Index](https://issues.openmrs.org/browse/OCLM-67)

### Summary
When performing a concept import from the API, we convert the fully specified name types from the api into the formats used by OpenMRS. This is implemented for the fully specified name type, but not for the Index Term and Short name types.


HerbertYiga:
could this ticket be having a link,cannot find it in the ticket description!!

Karuhanga:
> could this ticket be having a link,cannot find it in the ticket description!!

Thanks for pointing this out @HerbertYiga. Updated the description accordingly.

PR #59, opened 2019-12-09
Added ability to subscribe to a specific version of a concept collection

jecihjoy:
**Description of what I changed**

I added ability to subscribe to a specific version of a concept collection

**Issue I worked on**

https://issues.openmrs.org/browse/OCLM-74

**Checklist: I completed these to help reviewers :)**

- [x] My pull request only contains ONE single commit
(the number above, next to the 'Commits' tab is 1).

   No? -> [read here](https://wiki.openmrs.org/display/docs/Pull+Request+Tips) on how to squash multiple commits into one
- [x] My IDE is configured to follow the [code style](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with git add . && git commit --amend

- [x]  I have added tests to cover my changes. (If you refactored
existing code that was well tested you do not have to add tests)

   No? -> write tests and add them to this commit git add . && git commit --amend
- [x]  I ran mvn clean package right before creating this pull request and
added all formatting changes to my commit.

    No? -> execute above command
- [x] All new and existing tests passed.

   No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.
- [x] My pull request is based on the latest changes of the master branch.

   No? Unsure? -> execute command git pull --rebase upstream master

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
@@ -28,6 +28,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.URL;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -42,11 +43,13 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
 public class OclClient {
-	
+
 	public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-	
+
 	public static final String FILE_NAME_FORMAT = "yyyyMMdd_HHmmss";
-	
+
+	public static final int NUMBER_OF_SLASHES_AFTER_BASE_URL = 5;
+
 	private final String dataDirectory;
 	
 	private final int bufferSize = 64 * 1024;
@@ -56,7 +59,7 @@ public class OclClient {
 	private volatile long bytesDownloaded = 0;
 	
 	private volatile long totalBytesToDownload = 0;
-	
+
 	public OclClient() {
 		dataDirectory = OpenmrsUtil.getApplicationDataDirectory();
 	}
@@ -64,7 +67,7 @@ public class OclClient {
 	public OclClient(String dataDirectory) {
 		this.dataDirectory = dataDirectory;
 	}
-	
+
 	public OclResponse fetchSnapshotUpdates(String url, String token, Date updatedSince) throws IOException {
 		totalBytesToDownload = -1; //unknown yet
 		bytesDownloaded = 0;
@@ -99,20 +102,20 @@ public class OclClient {
 		return extractResponse(get);
 	}
 	
-	public OclResponse fetchLastReleaseVersion(String url, String token) throws IOException {
+	public OclResponse fetchOclConcepts(String url, String token) throws IOException {
 		totalBytesToDownload = -1; //unknown yet
 		bytesDownloaded = 0;
 
-		String latestVersion = fetchLatestOclReleaseVersion(url, token);
+		String collectionVersion = getOclReleaseVersion(url, token);
 
-		GetMethod exportUrlGet = executeExportRequest(url, latestVersion);
+		GetMethod exportUrlGet = executeExportRequest(url, collectionVersion);
 		
 		return extractResponse(exportUrlGet);
     }
 
-    public GetMethod executeExportRequest(String url, String latestVersion) throws IOException{
+    public GetMethod executeExportRequest(String url, String collectionVersion) throws IOException{
 
-		GetMethod exportUrlGet = new GetMethod(url + "/" + latestVersion + "/export");
+		GetMethod exportUrlGet = getExportUrl(url, collectionVersion);
 
 		HttpClient client = new HttpClient();
 		client.getHttpConnectionManager().getParams().setSoTimeout(TIMEOUT_IN_MS);
@@ -125,13 +128,73 @@ public class OclClient {
 		return exportUrlGet;
 	}
 
-	public OclResponse fetchLastReleaseVersion(String url, String token, String lastReleaseVersion) throws IOException {
-		String latestOclReleaseVersion = fetchLatestOclReleaseVersion(url, token);
-		if (lastReleaseVersion == null || !lastReleaseVersion.equals(latestOclReleaseVersion)) {
+	/**
+	 * This constructs the url to import concepts from the subscription url provided
+	 * @param url, the subscription url
+	 * @param version, desired version of the specified collection
+	 * @return GetMethod configured with the export url
+	 */
+	public GetMethod getExportUrl(String url, String version){
+		GetMethod exportUrlGet;
+
+		url = removeLastUrlForwardSlashIfExist(url);
+
+		if(url.contains(version)){
+			exportUrlGet = new GetMethod(url + "/export");
+		}else {
+			exportUrlGet = new GetMethod(url + "/" + version + "/export");
+		}
+		return exportUrlGet;
+	}
+
+	/**
+	 * This gets the desired collection release version
+	 * @param url, the subscription url
+	 * @param token, the subscription API token
+	 * @return the retrieved collection version
+	 * @throws IOException
+	 */
+	public String  getOclReleaseVersion(String url, String token) throws IOException {
+		String exportVersion;
+		url = removeLastUrlForwardSlashIfExist(url);
+		URL subscriptionURL = new URL(url);
+
+		String subUrlAfterBaseUrl = subscriptionURL.getPath();
+
+
+		int count = StringUtils.countMatches(subUrlAfterBaseUrl, "/");
+		/**
+		 *This checks if collection version has been passed to subscription url by checking number of forward slashes after ocl base url
+		 *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
+		 *that means collection version was passed and it's assigned to exportVersion
+		 */
+		if (count == NUMBER_OF_SLASHES_AFTER_BASE_URL){
+			exportVersion =  url.substring(url.lastIndexOf("/") + 1);
+		}else {
+			exportVersion = fetchLatestOclReleaseVersion(url, token);
+		}
+		return exportVersion;
+	}
+
+	/**
+	 * This changes the url by removing the last forward slash if it exists
+	 * @param url, this is the subscription url
+	 * @return the url format expected by the ocl server
+	 */
+	public String removeLastUrlForwardSlashIfExist(String url){
+		if (url.endsWith("/")) {
+			url = url.substring(0, url.lastIndexOf('/'));
+		}
+		return url;
+	}
+
+	public OclResponse fetchOclConcepts(String url, String token, String lastReleaseVersion) throws IOException {
+		String versionToImport = getOclReleaseVersion(url, token);
+		if (lastReleaseVersion == null || !lastReleaseVersion.equals(versionToImport)) {
 			//If there is no lastReleaseVersion then the subscription has been changed from snapshot to releases
 			//and we need to fetch the latest OCL release version.
 			//If lastReleaseVersion does not match the latest OCL release version then we need to fetch it too.
-			return fetchLastReleaseVersion(url, token);
+			return fetchOclConcepts(url, token);
 		}
 		else {
 			//No new version
@@ -327,9 +390,7 @@ public class OclClient {
     }
 
 	public String fetchLatestOclReleaseVersion(String url, String token) throws IOException {
-		if (url.endsWith("/")) {
-			url = url.substring(0, url.lastIndexOf('/'));
-		}
+		url = removeLastUrlForwardSlashIfExist(url);
 
 	    String latestVersionUrl = url + "/versions";
 		
--- a/api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
+++ b/api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
@@ -10,6 +10,8 @@
 package org.openmrs.module.openconceptlab.client;
 
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.startsWith;
 import static org.junit.Assert.assertThat;
@@ -22,28 +24,35 @@ import java.util.Calendar;
 import java.util.Date;
 
 import org.apache.commons.httpclient.Header;
+import org.apache.commons.httpclient.URIException;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.httpclient.util.DateUtil;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
 import org.openmrs.module.openconceptlab.MockTest;
 import org.openmrs.module.openconceptlab.TestResources;
-import org.openmrs.module.openconceptlab.client.OclClient;
 import org.openmrs.module.openconceptlab.client.OclClient.OclResponse;
 
 public class OclClientTest extends MockTest {
-	
+
 	@Mock
 	GetMethod get;
-	
+
 	OclClient oclClient;
-	
+
 	File tempDir;
-	
+
+	private static final String URL = "https://api.openconceptlab.org/users/username/collections/collectionname/";
+
+	private static final String URL_WITH_VERSION = "https://api.openconceptlab.org/users/username/collections/collectionname/v1.0";
+
+	private static final String URL_WITHOUT_VERSION = "https://api.openconceptlab.org/users/username/collections/collectionname";
+
 	@Before
 	public void createTempDir() throws IOException {
 		tempDir = File.createTempFile("ocl", "");
@@ -131,5 +140,36 @@ public class OclClientTest extends MockTest {
 		calendar.set(2015, 5, 22, 12, 12, 29);
 		
 		assertThat(date, is(calendar.getTime()));
+	}
+
+	@Test
+	public void getOclReleaseVersion_shouldGetOclReleaseVersion() throws IOException {
+		String subscriptionToken = "53fc72f0498a707a26e4d903c0f24c2db24d1e35";
+		String version = oclClient.getOclReleaseVersion(URL_WITH_VERSION, subscriptionToken);
+		assertThat(version, is("v1.0"));
+	}
+
+	@Test
+	public void getExportUrl_shouldConstructCorrectExportUrlWhenVersionIsPassedToUrl() throws URIException {
+		String collectionVersion = "v1.0";
+		GetMethod getMethodUrlWithVersion = oclClient.getExportUrl(URL_WITH_VERSION, collectionVersion);
+		assertThat(getMethodUrlWithVersion.getURI().toString(), is(URL_WITH_VERSION + "/export"));
+	}
+
+	@Test
+	public void getExportUrl_shouldConstructCorrectExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String collectionVersion = "v1.0";
+		GetMethod getMethodUrlWithoutVersion = oclClient.getExportUrl(URL_WITHOUT_VERSION, collectionVersion);
+		assertThat(getMethodUrlWithoutVersion.getURI().toString(), is(URL_WITHOUT_VERSION + "/v1.0" + "/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldRemoveLastForwardSlashIfItExist() {
+		String newUrl = oclClient.removeLastUrlForwardSlashIfExist(URL);
+		int numberOfForwardSlashesInUrl = StringUtils.countMatches(URL, "/");
+
+		int numberOfForwardSlashesInNewUrl = StringUtils.countMatches(newUrl, "/");
+
+		assertThat(numberOfForwardSlashesInUrl, equalTo(numberOfForwardSlashesInNewUrl + 1));
 	}
 }
--- a/omod/src/main/resources/config.xml
+++ b/omod/src/main/resources/config.xml
@@ -54,7 +54,6 @@
 		<description>The OCL API Token</description>
 	</globalProperty>
 
-
 	<!-- Maps hibernate file's, if present -->
 	<mappingFiles>
 				
--- a/owa/app/js/subscription/subscription.controller.js
+++ b/owa/app/js/subscription/subscription.controller.js
@@ -5,10 +5,12 @@ class SubscriptionController {
     $rootScope.links["Open Concept Lab"] = "";
     $rootScope.links["Subscription"] = "subscription";
 
+    const NUMBER_OF_SLASHES_AFTER_BASE_URL = 5;
     var vm = this;
     vm.cancel = cancel;
     vm.subscribe = subscribe;
     vm.unSubscribe = unSubscribe;
+    vm.isVersionAdded = isVersionAdded;
 
     function cancel(){
       $location.path('/');
@@ -47,6 +49,27 @@ class SubscriptionController {
       openmrsRest.getFull("openconceptlab/subscription").then(function (response) {
         vm.subscription = response.results[0];
       })
+    }
+
+    /*
+    *This checks if collection version has been passed to subscription url by checking number of forward slashes after base url
+    *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
+    *that means collection version was passed and isVersionAdded() will return true
+    */
+    function isVersionAdded(subscriptionUrl) {
+    if (subscriptionUrl.endsWith("/")) {
+            subscriptionUrl = subscriptionUrl.substring(0, subscriptionUrl.lastIndexOf('/'));
+    }
+    var url = new URL(subscriptionUrl);
+    let subUrlAfterBaseUrl = url.pathname;
+
+      let count = subUrlAfterBaseUrl.length - subUrlAfterBaseUrl.replace(/[\/\\]/g, '').length;
+      if (count == NUMBER_OF_SLASHES_AFTER_BASE_URL) {
+        vm.subscription.subscribedToSnapshot = false;
+        return true;
+      } else {
+        return false
+      }
     }
   }
 }
--- a/owa/app/js/subscription/subscription.html
+++ b/owa/app/js/subscription/subscription.html
@@ -13,8 +13,8 @@
                 </p>
                 <p><a ng-click="vm.isShowAdvanced = !vm.isShowAdvanced" ng-init="vm.isShowAdvanced = false">{{"showAdvanced.text" | translate}}</a></p>
                 <div ng-show="vm.isShowAdvanced">
-                    <input ng-model="vm.subscription.subscribedToSnapshot" type="checkbox"> {{"subscription.options.checkbox" | translate}}<br>
-                    <input ng-model="vm.subscription.validationType" type="checkbox" ng-true-value="'NONE'" ng-false-value="'FULL'"> {{"subscription.validationType.checkbox" | translate}}<br>
+                    <input ng-disabled="vm.isVersionAdded(vm.subscription.url)" ng-model="vm.subscription.subscribedToSnapshot" type="checkbox"> {{"subscription.options.checkbox" | translate}}<br>
+                    <p style="color:#ffcc00; font-size:12px" ng-show="vm.isVersionAdded(vm.subscription.url)">You cannot subscribe to a SNAPSHOT if you've provided the collection version</p>                    <input ng-model="vm.subscription.validationType" type="checkbox" ng-true-value="'NONE'" ng-false-value="'FULL'"> {{"subscription.validationType.checkbox" | translate}}<br>
                 </div>
             </div>
             <br>
--- a/api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
@@ -152,18 +152,18 @@ public class Importer implements Runnable {
                 OclResponse oclResponse;
 
                 if (updatedSince == null) {
-                    oclResponse = oclClient.fetchLastReleaseVersion(subscription.getUrl(), subscription.getToken());
+                    oclResponse = oclClient.fetchOclConcepts(subscription.getUrl(), subscription.getToken());
 					importService.updateReleaseVersion(anImport,
-                            oclClient.fetchLatestOclReleaseVersion(subscription.getUrl(), subscription.getToken()));
+                            oclClient.getOclReleaseVersion(subscription.getUrl(), subscription.getToken()));
                 } else {
                     if (subscription.isSubscribedToSnapshot()) {
                         oclResponse = oclClient.fetchSnapshotUpdates(subscription.getUrl(), subscription.getToken(),
                                 updatedSince);
                     }
                     else {
-                        oclResponse = oclClient.fetchLastReleaseVersion(subscription.getUrl(), subscription.getToken(), lastImport.getReleaseVersion());
+                        oclResponse = oclClient.fetchOclConcepts(subscription.getUrl(), subscription.getToken(), lastImport.getReleaseVersion());
 						importService.updateReleaseVersion(anImport,
-                                oclClient.fetchLatestOclReleaseVersion(subscription.getUrl(), subscription.getToken()));
+                                oclClient.getOclReleaseVersion(subscription.getUrl(), subscription.getToken()));
                     }
                 }
 
--- a/api/src/test/java/org/openmrs/module/openconceptlab/ImportServiceTest.java
+++ b/api/src/test/java/org/openmrs/module/openconceptlab/ImportServiceTest.java
@@ -331,7 +331,7 @@ public class ImportServiceTest extends BaseModuleContextSensitiveTest {
 			when(mockedGetMethod.getResponseContentLength()).thenReturn(Long.valueOf(response.available()));
 
 			when(mockedOclClient.executeExportRequest(importService.getSubscription().getUrl(), version)).thenReturn(mockedGetMethod);
-			when(mockedOclClient.fetchLastReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
+			when(mockedOclClient.fetchOclConcepts(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
 			when(mockedOclClient.fetchLatestOclReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenReturn(version);
 
 			SimpleDateFormat format = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z", Locale.US);
@@ -405,7 +405,7 @@ public class ImportServiceTest extends BaseModuleContextSensitiveTest {
 			when(mockedGetMethod.getResponseContentLength()).thenReturn(Long.valueOf(initialResponse.available()));
 
 			when(mockedOclClient.executeExportRequest(importService.getSubscription().getUrl(), initialVersion)).thenReturn(mockedGetMethod);
-			when(mockedOclClient.fetchLastReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
+			when(mockedOclClient.fetchOclConcepts(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
 			when(mockedOclClient.fetchLatestOclReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenReturn(initialVersion);
 
 			SimpleDateFormat format = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z", Locale.US);
@@ -439,7 +439,7 @@ public class ImportServiceTest extends BaseModuleContextSensitiveTest {
 			when(mockedGetMethod.getResponseContentLength()).thenReturn(Long.valueOf(followupResponse.available()));
 
 			when(mockedOclClient.executeExportRequest(importService.getSubscription().getUrl(), followupVersion)).thenReturn(mockedGetMethod);
-			when(mockedOclClient.fetchLastReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken(), initialVersion)).thenCallRealMethod();
+			when(mockedOclClient.fetchOclConcepts(importService.getSubscription().getUrl(),importService.getSubscription().getToken(), initialVersion)).thenCallRealMethod();
 			when(mockedOclClient.fetchLatestOclReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenReturn(followupVersion);
 
 			date = format.parse(followupDateHeaderValue);
@@ -503,7 +503,7 @@ public class ImportServiceTest extends BaseModuleContextSensitiveTest {
 			when(mockedGetMethod.getResponseContentLength()).thenReturn(Long.valueOf(response.available()));
 
 			when(mockedOclClient.executeExportRequest(importService.getSubscription().getUrl(), version)).thenReturn(mockedGetMethod);
-			when(mockedOclClient.fetchLastReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
+			when(mockedOclClient.fetchOclConcepts(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenCallRealMethod();
 			when(mockedOclClient.fetchLatestOclReleaseVersion(importService.getSubscription().getUrl(),importService.getSubscription().getToken())).thenReturn(version);
 
 			SimpleDateFormat format = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z", Locale.US);
--- a/api/src/test/java/org/openmrs/module/openconceptlab/importer/ImporterTest.java
+++ b/api/src/test/java/org/openmrs/module/openconceptlab/importer/ImporterTest.java
@@ -84,7 +84,7 @@ public class ImporterTest extends BaseContextMockTest {
 		Date updatedTo = new Date();
 		OclResponse oclResponse = new OclClient.OclResponse(IOUtils.toInputStream("{}"), 0, updatedTo);
 		when(importService.getLastImport()).thenReturn(null);
-		when(oclClient.fetchLastReleaseVersion(subscription.getUrl(), subscription.getToken())).thenReturn(oclResponse);
+		when(oclClient.fetchOclConcepts(subscription.getUrl(), subscription.getToken())).thenReturn(oclResponse);
 
 		importer.run();
 
@@ -115,7 +115,7 @@ public class ImporterTest extends BaseContextMockTest {
         OclResponse oclResponse = new OclClient.OclResponse(IOUtils.toInputStream("{}"), 0, new Date());
 
         when(oclClient.fetchLatestOclReleaseVersion(subscription.getUrl(), subscription.getToken())).thenReturn(release2name);
-        when(oclClient.fetchLastReleaseVersion(subscription.getUrl(), subscription.getToken(), lastImport.getReleaseVersion())).thenReturn(oclResponse);
+        when(oclClient.fetchOclConcepts(subscription.getUrl(), subscription.getToken(), lastImport.getReleaseVersion())).thenReturn(oclResponse);
 
         importer.run();
     }


dkayiwa:
Did you get a chance to look at this? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

jecihjoy:
I updated the PR

dkayiwa:
@jecihjoy can you include the ticket id in the commit message as recommended on the above pull request tips page?

jecihjoy:
That's done @dkayiwa @rkorytkowski

dkayiwa:
@jecihjoy did you take a look at the travis build failure?

dkayiwa:
With these changes are we still able to subscribe to the latest version as before?

jecihjoy:
Yes by 
1. Not proving collection version in the subscription url
2. Indicating the latest version in the url

dkayiwa:
And the travis build failure?

jecihjoy:
I did that in a different PR
 https://github.com/openmrs/openmrs-module-openconceptlab/pull/60

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +49,22 @@ class SubscriptionController {
         vm.subscription = response.results[0];
       })
     }
+
+    function isVersionAdded(url) {
+      if (url.endsWith("/")) {
+        url = url.substring(0, url.lastIndexOf('/'));
+      }
+      let words = url.split("api.openconceptlab.org")
```
dkayiwa: Are we intentionally hard coding `api.openconceptlab.org`?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +132,13 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
 		
 		assertThat(date, is(calendar.getTime()));
 	}
+
+	@Test
+	public void getUrlVersion_shouldGetUrlVersion() throws IOException {
```
dkayiwa: Can we add another test for what the expected return value should be for cases dealing with the latest released version?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -103,16 +103,22 @@ public OclResponse fetchLastReleaseVersion(String url, String token) throws IOEx
...
 		
 		return extractResponse(exportUrlGet);
     }
 
     public GetMethod executeExportRequest(String url, String latestVersion) throws IOException{
 
-		GetMethod exportUrlGet = new GetMethod(url + "/" + latestVersion + "/export");
+		GetMethod exportUrlGet;
+
+		if(url.contains(latestVersion)){
```
dkayiwa: Can you create a method for this and add a test for it?

owa/app/js/subscription/subscription.controller.js
jecihjoy: So apparently the version section of the URL is not passed as a URL Param and hence extracting it's value it's a little tricky that's why I went the string processing way
However I'm open to suggestions on how to go about this @dkayiwa

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +49,22 @@ class SubscriptionController {
...
+
+      let count = words[1].length - words[1].replace(/[\/\\]/g, '').length;
+
+      if (count == 5) {
```
dkayiwa: What is the meaning of the magic number `5`?

owa/app/js/subscription/subscription.html
```diff
@@ -13,8 +13,8 @@
                 </p>
                 <p><a ng-click="vm.isShowAdvanced = !vm.isShowAdvanced" ng-init="vm.isShowAdvanced = false">{{"showAdvanced.text" | translate}}</a></p>
                 <div ng-show="vm.isShowAdvanced">
-                    <input ng-model="vm.subscription.subscribedToSnapshot" type="checkbox"> {{"subscription.options.checkbox" | translate}}<br>
-                    <input ng-model="vm.subscription.validationType" type="checkbox" ng-true-value="'NONE'" ng-false-value="'FULL'"> {{"subscription.validationType.checkbox" | translate}}<br>
+                    <input ng-disabled="vm.isVersionAdded(vm.subscription.url)" ng-model="vm.subscription.subscribedToSnapshot" type="checkbox"> {{"subscription.options.checkbox" | translate}}<br>
+                    <p style="color:#ffcc00; font-size:12px" ng-show="vm.isVersionAdded(vm.subscription.url)">You cannot subscript to a SNAPSHOT if you've provided the collection version</p>                    <input ng-model="vm.subscription.validationType" type="checkbox" ng-true-value="'NONE'" ng-false-value="'FULL'"> {{"subscription.validationType.checkbox" | translate}}<br>
```
dkayiwa: subscript or subscribe?

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +49,22 @@ class SubscriptionController {
         vm.subscription = response.results[0];
       })
     }
+
+    function isVersionAdded(url) {
```
dkayiwa: What is this function trying to do?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +131,26 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+		String exportVersion;
+
+		if (url.endsWith("/")) {
+			url = url.substring(0, url.lastIndexOf('/'));
+		}
+		String[] words = url.split("api.openconceptlab.org");
+
+		int count = words[1].length() - words[1].replaceAll("/","").length();
+
+		if (count == 5){
```
dkayiwa: What is the meaning of the magic number `5`?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +131,26 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
 		return exportUrlGet;
 	}
 
+	public String  getUrlVersion(String url, String token) throws IOException {
+		String exportVersion;
+
+		if (url.endsWith("/")) {
+			url = url.substring(0, url.lastIndexOf('/'));
+		}
+		String[] words = url.split("api.openconceptlab.org");
```
dkayiwa: Instead of hard coding `api.openconceptlab.org`, how about putting it in a global property?

owa/app/js/subscription/subscription.controller.js
jecihjoy: I'm using that while doing the string manipulations too, maybe I should have it declared someone, will that make any difference?
I mean not having the version URL section presented as a Param introduces all these codes

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: Noted

owa/app/js/subscription/subscription.controller.js
dkayiwa: A constant with a name which help me know why it is not 6 or 7, would be helpful. 😊

jecihjoy:
I just added some comments here https://issues.openmrs.org/browse/OCLM-74

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -42,11 +45,16 @@
...
 public class OclClient {
-	
+
 	public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-	
+
 	public static final String FILE_NAME_FORMAT = "yyyyMMdd_HHmmss";
-	
+
+	public static final String oclApiUrl = "api.openconceptlab.org";
```
dkayiwa: Did you look at our naming convention for constants?https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-NamingConventions

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -42,11 +45,16 @@
...
+
+	//if number is five that means the subscription url contains the collection version
+	public static final int NUMBER_OF_FORWARD_SLASHES = 5;
```
dkayiwa: Can you give an example of what such a value would look like with the 5  forward slashes?

jecihjoy: https://api.openconceptlab.org/users/Jecihjoy/collections/NOC/v2.0

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +133,45 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
 		return exportUrlGet;
 	}
 
+	public GetMethod getExportUrl(String url, String latestVersion){
```
dkayiwa: Can you add some javadoc to these methods?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +133,45 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+
+		if(url.contains(latestVersion)){
+			exportUrlGet = new GetMethod(url + "/export");
+		}else {
+			exportUrlGet = new GetMethod(url + "/" + latestVersion + "/export");
+		}
+		return exportUrlGet;
+	}
+
+	public String  getUrlVersion(String url, String token) throws IOException {
```
dkayiwa: Can you add some javadoc to these methods?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +133,45 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+
+		if (count == NUMBER_OF_FORWARD_SLASHES){
+			exportVersion =  url.substring(url.lastIndexOf("/") + 1); //return url version
+		}else {
+			exportVersion = fetchLatestOclReleaseVersion(url, token);
+		}
+		return exportVersion;
+	}
+
+	public String removeLastUrlForwardSlashIfExist(String url){
```
dkayiwa: Can you add some javadoc to these methods?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: Like in code or just here in the comments?

owa/app/js/subscription/subscription.controller.js
```diff
@@ -5,10 +5,13 @@ class SubscriptionController {
     $rootScope.links["Open Concept Lab"] = "";
     $rootScope.links["Subscription"] = "subscription";
 
+    var oclApiUrl = "api.openconceptlab.org"
```
dkayiwa: Are these constants or variables that are to be changed?

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +51,23 @@ class SubscriptionController {
         vm.subscription = response.results[0];
       })
     }
+
+    //Check if collection version has been passed to subscription url by checking number of forward slashes
```
dkayiwa: Can you give an example of what that would look like?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +133,37 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	public void getExportUrl_shouldConstructCorretExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String url = "https://api.openconceptlab.org/users/username/collections/collectionname/";
+		String collectionVersion = "v2.0";
+		GetMethod getMethod = oclClient.getExportUrl(url, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(url + "v2.0/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldWorkAsExpected() {
+		String url1 = "https://api.openconceptlab.org/users/username/collections/collectionname/";
```
dkayiwa: Are the urls variables to be changed? Or constants?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +133,37 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+		String url2 = "https://api.openconceptlab.org/users/username/collections/collectionname";
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(url1);
+		String newUrl2 = oclClient.removeLastUrlForwardSlashIfExist(url2);
```
dkayiwa: The returned values being the same does not necessarily mean that the last forward slash has been removed. For instance, if some maintainer accidentally introduces a bug and the two method calls return `https://api.openconceptlab.org/users/username/collections/collectionname/`, this test will still pass.

jecihjoy: I'll add more check to cater for that

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: They are to be changed, the method removes the last '/' at the end of the url if it exists

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
dkayiwa: Which one do you think would be better? 😊

owa/app/js/subscription/subscription.controller.js
jecihjoy: In the comments or in the code?

url without version (4 forward slashes)
https://api.openconceptlab.org/users/username/collections/collectionname

url with version (5 forward slashes)
https://api.openconceptlab.org/users/username/collections/collectionname/v1.0

5 forward slashes means you have provided the collection version and hence you can't subscribe to a snapshot

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
dkayiwa: In this test method, i do not see any where you are changing the variables url1 and url2

owa/app/js/subscription/subscription.controller.js
dkayiwa: Which one do you think is better?

owa/app/js/subscription/subscription.controller.js
jecihjoy: constant, for variable I would have used [let](https://www.w3schools.com/js/js_let.asp) keyword

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: This method "removeLastUrlForwardSlashIfExist" is converting url1 to https://api.openconceptlab.org/users/username/collections/collectionname

and url2 remain the same since it doesn't end with '/'

dkayiwa: What is mean is that as far as the test method is concerned, they are constants. Correct?

jecihjoy: Yes I get it now converting them to constants..

owa/app/js/subscription/subscription.controller.js
jecihjoy: I changed this value into a constant

owa/app/js/subscription/subscription.html
jecihjoy: Fixed

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: Change oclApiUrl to OCL_API_URL

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: @dkayiwa  I'm still not sure about this, does it make sense to have a lot of comments in the code?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: done

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: done!

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: Done

owa/app/js/subscription/subscription.controller.js
jecihjoy: @dkayiwa  I'm still not sure about this, does it make sense to have a lot of comments in the code?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
dkayiwa: What is your definition of `a lot`?

owa/app/js/subscription/subscription.controller.js
dkayiwa: How much is a lot?

jecihjoy: I came up with
    /*
    *This checks if collection version has been passed to subscription url by checking number of forward slashes
    *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
    *that means collection version was passed and isVersionAdded() will return true
    */

hopefully it's good enough?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: Is this good enough?

		/*
		 *This checks if collection version has been passed to subscription url by checking number of forward slashes
		 *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
		 *that means collection version was passed and it's assigned to exportVersion
		 */

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +132,68 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+
+		url = removeLastUrlForwardSlashIfExist(url);
+
+		String[] words = url.split(OCL_API_URL);
+
+		int count = words[1].length() - words[1].replaceAll("/","").length();
+
+		/*
+		 *This checks if collection version has been passed to subscription url by checking number of forward slashes
+		 *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
```
dkayiwa: Is the number 5 or 7?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +132,68 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
 		return exportUrlGet;
 	}
 
+	/**
+	 * This constructs the url to import concepts from the subscription url provided
+	 * @param url
```
dkayiwa: Can you document the params too?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +132,68 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+			exportUrlGet = new GetMethod(url + "/export");
+		}else {
+			exportUrlGet = new GetMethod(url + "/" + latestVersion + "/export");
+		}
+		return exportUrlGet;
+	}
+
+	/**
+	 * This gets the desired collection release version
+	 * @param url
```
dkayiwa: Can you document the params too?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +132,68 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+			exportVersion =  url.substring(url.lastIndexOf("/") + 1); //return url version
+		}else {
+			exportVersion = fetchLatestOclReleaseVersion(url, token);
+		}
+		return exportVersion;
+	}
+
+	/**
+	 * This changes the url by removing the last forward slash if it exists
+	 * @param url
```
dkayiwa: Can you document the params too?

owa/app/js/subscription/subscription.controller.js
dkayiwa: Is the above a constant?

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +51,27 @@ class SubscriptionController {
         vm.subscription = response.results[0];
       })
     }
+
+    /*
+    *This checks if collection version has been passed to subscription url by checking number of forward slashes
+    *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
```
dkayiwa: Is the number 5 or 7?

owa/app/js/subscription/subscription.controller.js
jecihjoy: It's a constant.
for variable I would have used let keyword

owa/app/js/subscription/subscription.controller.js
jecihjoy: I already documented this

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: I already documented this

owa/app/js/subscription/subscription.controller.js
dkayiwa: What i mean is that the number is not 5 in the example you have given.

owa/app/js/subscription/subscription.controller.js
dkayiwa: If i later on assign a value to `oclApiUrl`, will the compiler complain?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,8 +132,68 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+	}
+
+	/**
+	 * This gets the desired collection release version
+	 * @param url the subscription url
+	 * @param token the subscription API token
+	 * @return the retrieved collection version
+	 * @throws IOException
+	 */
+	public String  getOclReleaseVersion(String url, String token) throws IOException {
```
dkayiwa: Where is the IOException thrown in this method?

jecihjoy: https://github.com/openmrs/openmrs-module-openconceptlab/blob/98eb2764b6fd564fd6be40daae19a8f12f0544c8/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java#L178

owa/app/js/subscription/subscription.controller.js
jecihjoy: I have update the descriptions

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: done

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: done

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: done

owa/app/js/subscription/subscription.controller.js
```diff
@@ -5,10 +5,13 @@ class SubscriptionController {
     $rootScope.links["Open Concept Lab"] = "";
     $rootScope.links["Subscription"] = "subscription";
 
+    const oclApiUrl = "api.openconceptlab.org"
```
dkayiwa: Did you take a look at our naming convention for constants?

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +51,27 @@ class SubscriptionController {
...
     }
+
+    /*
+    *This checks if collection version has been passed to subscription url by checking number of forward slashes after https://api.openconceptlab.org
+    *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
+    *that means collection version was passed and isVersionAdded() will return true
+    */
+    function isVersionAdded(subscriptionUrl) {
+      if (subscriptionUrl.endsWith("/")) {
+        subscriptionUrl = subscriptionUrl.substring(0, url.lastIndexOf('/'));
```
dkayiwa: Should it be `url.lastIndexOf` or `subscriptionUrl.lastIndexOf`

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +51,27 @@ class SubscriptionController {
...
+      }
+      let words = subscriptionUrl.split(oclApiUrl)
+
+      let count = words[1].length - words[1].replace(/[\/\\]/g, '').length;
+
+      if (count == numberOfForwardSlashes) {
```
dkayiwa: What does numberOfForwardSlashes mean?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -42,11 +45,15 @@
...
 public class OclClient {
-	
+
 	public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-	
+
 	public static final String FILE_NAME_FORMAT = "yyyyMMdd_HHmmss";
-	
+
+	public static final String OCL_API_URL = "api.openconceptlab.org";
```
dkayiwa: OCL_API__BASE_URL

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -42,11 +45,16 @@
...
 	public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-	
+
 	public static final String FILE_NAME_FORMAT = "yyyyMMdd_HHmmss";
-	
+
+	public static final String OCL_API_BASE_URL = "api.openconceptlab.org";
+
+	//number of special character '/' in the subscription url after the ocl api base url
+	public static final int NUMBER_OF_FORWARD_SLASHES = 5;
```
dkayiwa: NUMBER_OF_SLASHES_AFTER_BASE_URL

owa/app/js/subscription/subscription.controller.js
```diff
@@ -5,10 +5,14 @@ class SubscriptionController {
     $rootScope.links["Open Concept Lab"] = "";
     $rootScope.links["Subscription"] = "subscription";
 
+    const OCL_API_BASE_URL = "api.openconceptlab.org"
+    //number of special character '/' in the subscription url after the ocl api base url
+    const NUMBER_OF_FORWARD_SLASHES = 5;
```
dkayiwa: NUMBER_OF_SLASHES_AFTER_BASE_URL

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +52,27 @@ class SubscriptionController {
...
+    /*
+    *This checks if collection version has been passed to subscription url by checking number of forward slashes after https://api.openconceptlab.org
+    *If the number is 5, such as with https://api.openconceptlab.org/users/username/collections/collectionname/v1.0
+    *that means collection version was passed and isVersionAdded() will return true
+    */
+    function isVersionAdded(subscriptionUrl) {
+      if (subscriptionUrl.endsWith("/")) {
+        subscriptionUrl = subscriptionUrl.substring(0, subscriptionUrl.lastIndexOf('/'));
+      }
+      let words = subscriptionUrl.split(OCL_API_BASE_URL)
```
dkayiwa: On the above line, are you trying to get a list of tokens where OCL_API_BASE_URL is the separator? Or are you simply trying to get the string after OCL_API_BASE_URL? If the latter, then i find the use of `split` confusing.

owa/app/js/subscription/subscription.controller.js
```diff
@@ -48,6 +52,27 @@ class SubscriptionController {
...
+
+      let count = words[1].length - words[1].replace(/[\/\\]/g, '').length;
```
dkayiwa: What is the meaning the magic number `1` in the index for words?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +143,40 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+		String collectionVersion = "v2.0";
+		GetMethod getMethod = oclClient.getExportUrl(URL_1, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(URL_1 + "v2.0/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldWorkAsExpected() {
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(URL_1);
+		String newUrl2 = oclClient.removeLastUrlForwardSlashIfExist(URL_2);
+		String[] url1Chars = URL_1.split(OCL_API_BASE_URL);
```
dkayiwa: My intuition of the split function is for a list of tokens separated by a separator. OCL_API_BASE_URL does not naturally look like a separator. It is instead look like a prefix for the url. So using split in this case does not make the code obvious to understand.

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +143,40 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+		int url1Count = url1Chars[1].length() - url1Chars[1].replaceAll("/","").length();
+
+		String[] newUrl1Chars = newUrl1.split(OCL_API_BASE_URL);
+		int newUrl1Count = newUrl1Chars[1].length() - newUrl1Chars[1].replaceAll("/","").length();
```
dkayiwa: Why index `1` instead of `0` in newUrl1Chars?

owa/app/js/subscription/subscription.controller.js
jecihjoy: Had tried using subString function before but let to more confusing code like 

w = word.substring(word.lastIndexOf('org') + 3) 

Do you have a better solution for this

dkayiwa: Ok tell me what you are trying to do with subscriptionUrl.split(OCL_API_BASE_URL) and then i will advise you accordingly.

jecihjoy: To get the sub string after OCL_API_BASE_URL

dkayiwa: Wouldn't this work?

subscriptionUrl.substring(OCL_API_BASE_URL.length);

jecihjoy: Actually that is return the whole string as it is

jecihjoy: waoh thank you

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,13 +132,73 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+	 * @return the retrieved collection version
+	 * @throws IOException
+	 */
+	public String  getOclReleaseVersion(String url, String token) throws IOException {
+		String exportVersion;
+
+		url = removeLastUrlForwardSlashIfExist(url);
+
+		String subUrlAfterBaseUrl = url.substring(OCL_API_BASE_URL.length());
+
```
dkayiwa: If in all this you are simply trying to get the number of slashes, then wouldn't be both simpler and clearer if you used? StringUtils.countMatches()

pom.xml
```diff
@@ -147,6 +147,11 @@
 					</exclusion>
 				</exclusions>
 			</dependency>
+			<dependency>
+				<groupId>org.apache.commons</groupId>
```
dkayiwa: Are you sure we need this dependency?

jecihjoy: To use this StringUtils.countMatches, I needed it 
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html

dkayiwa: Did you get any error when you did not have this?

jecihjoy: I did update the code

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -42,11 +42,15 @@
...
 public class OclClient {
-	
+
 	public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
-	
+
 	public static final String FILE_NAME_FORMAT = "yyyyMMdd_HHmmss";
-	
+
+	public static final String OCL_API_BASE_URL = "api.openconceptlab.org";
```
dkayiwa: What happens when this is changed to something like `api.qa.openconceptlab.org`? https://github.com/openmrs/openmrs-ocl-client/search?q=api.openconceptlab.org&unscoped_q=api.openconceptlab.org

Do we have to recompile the module?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +144,41 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+		GetMethod getMethod = oclClient.getExportUrl(URL_1, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(URL_1 + "v2.0/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldWorkAsExpected() {
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(URL_1);
+		String newUrl2 = oclClient.removeLastUrlForwardSlashIfExist(URL_2);
+
+		String url1Chars = URL_1.substring(OCL_API_BASE_URL.length());
```
dkayiwa: I find this test complicated. I though it was supposed to be as simple as confirming that the last slash is removed?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +144,41 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	public void getExportUrl_shouldConstructCorrectExportUrlWhenVersionIsPassedToUrl() throws URIException {
+		String collectionVersion = "v1.0";
+		GetMethod getMethod = oclClient.getExportUrl(URL_WITH_VERSION, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(URL_WITH_VERSION + "/export"));
+	}
+
+	@Test
+	public void getExportUrl_shouldConstructCorretExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String collectionVersion = "v2.0";
+		GetMethod getMethod = oclClient.getExportUrl(URL_1, collectionVersion);
```
dkayiwa: Since the one used above is `URL_WITH_VERSION`, isn't the code easier to understand if the one used here is `URL_WITHOUT_VERSION`?

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
jecihjoy: Thinking about it, I see two options of working this out
1. Adding the server base url as a global variable
That means having a field above the subscription url where the use specifies the server url and default one being  this, **api.openconceptlab.org**
2. Extracting this base url from the subscription url by extracting the sub url before /users 
Which one would you advice?

dkayiwa: @jecihjoy which do you think is better?

jecihjoy: Another option to get the url part after base url is
URL url = new URL(subscriptionUrl);
subUrlAfterBaseUrl = url.getPath;

In this case we wouldn't care to know anything about the server base url.

I was thinking this is the best option, what is your option?

dkayiwa: I do not like a solution which complicates life for the end users. 😊

jecihjoy: However I have already saved the server base url as a global variable, if it makes sense I could retain the code but use the above getPath method.

dkayiwa: What i do not like is users having to break the url into two parts whenever they come to this screen. That is what i call a complicated user interface.

dkayiwa: So go with any solution where the user interface stays as it was, and change of the api server does not require us to recompile the module.

jecihjoy: That makes sense, actually I was not having the user provide the two url part as in the the screenshot I shared, I had gone the path of getting the host url form the subscription url provided and saving it as a global variable

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: I have added test cases for both

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: Simplified this

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -399,6 +399,15 @@ public void saveSubscription(Subscription subscription) {
 		url.setPropertyValue(prependApiIfAbsent(subscription.getUrl()));
 		adminService.saveGlobalProperty(url);
 
+		GlobalProperty apiUrl = new GlobalProperty(OpenConceptLabConstants.GP_OCL_API_URL);
+		String subscriptionUrl = adminService.getGlobalProperty(OpenConceptLabConstants.GP_SUBSCRIPTION_URL);
+		try {
+			apiUrl.setPropertyValue(new URL(subscriptionUrl).getHost());
```
dkayiwa: If you can always dynamically get the base url from whatever the user supplies, without hard coding it, then what is the use of the global property?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
dkayiwa: This is the method that i expected to use `URL_WITHOUT_VERSION`

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +145,41 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	public void getOclReleaseVersion_shouldGeOclReleaseVersion() throws IOException {
+		String subscriptionToken = "53fc72f0498a707a26e4d903c0f24c2db24d1e35";
+		String version = oclClient.getOclReleaseVersion(URL_WITH_VERSION, subscriptionToken);
+		assertThat(version, is("v1.0"));
+	}
+
+	@Test
+	public void getExportUrl_shouldConstructCorrectExportUrlWhenVersionIsPassedToUrl() throws URIException {
+		String collectionVersion = "v1.0";
+		GetMethod getMethodUrlWithVersion = oclClient.getExportUrl(URL_WITH_VERSION, collectionVersion);
```
dkayiwa: According to the method name, i would expect this method to use `URL_WITH_VERSION` but not `URL_WITHOUT_VERSION`

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +145,41 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	@Test
+	public void getExportUrl_shouldConstructCorretExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String collectionVersion = "v2.0";
+		GetMethod getMethod = oclClient.getExportUrl(URL_1, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(URL_1 + "v2.0/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldWorkAsExpected() {
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(URL_1);
```
dkayiwa: I think you are complicating this test more than you should.
I would expect it to be as simple as, when passed a url which ends with a slash, it removes it.
That should be all.

jecihjoy: I'm not sure I follow what you saying but most of the lines in this code are to help make the correct assertions.
Kindly point out which line is not necessary here..

dkayiwa: Ok let me put it differently. You could call removeLastUrlForwardSlashIfExist() just once, in this test.

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +141,43 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	public void getExportUrl_shouldConstructCorretExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String collectionVersion = "v2.0";
+		GetMethod getMethod = oclClient.getExportUrl(URL_1, collectionVersion);
+		assertThat(getMethod.getURI().toString(), is(URL_1 + "v2.0/export"));
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldRemoveLastForwardSlashIfItExist() {
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(URL_1);
+		int Url1Count = StringUtils.countMatches(URL_1, "/");
```
dkayiwa: Any reason for appending `1` to the new url variable? Same for count.

dkayiwa: Can you take a look at our naming conventions for variables?

omod/src/main/resources/config.xml
```diff
@@ -53,6 +53,11 @@
 		<property>openconceptlab.token</property>
 		<description>The OCL API Token</description>
 	</globalProperty>
+		<globalProperty>
+		<property>openconceptlab.oclApiUrl</property>
```
dkayiwa: What are we using this for?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -22,28 +24,35 @@
...
-	
+
 	OclClient oclClient;
-	
+
 	File tempDir;
-	
+
+	public static final String URL_1 = "https://api.openconceptlab.org/users/username/collections/collectionname/";
+
```
dkayiwa: Should these constants be public?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +141,43 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldRemoveLastForwardSlashIfItExist() {
+		String newUrl1 = oclClient.removeLastUrlForwardSlashIfExist(URL_1);
+		int Url1Count = StringUtils.countMatches(URL_1, "/");
+
+		int newUrl1Count = StringUtils.countMatches(newUrl1, "/");
+
+		assertNotEquals(Url1Count, newUrl1Count);
+		assertThat(Url1Count, greaterThan(newUrl1Count));
```
dkayiwa: Do we need these two assertions? In other words, if one is greater than the other, is there any possibility of being equal? 😊

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,13 +128,73 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+	 * This gets the desired collection release version
+	 * @param url, the subscription url
+	 * @param token, the subscription API token
+	 * @return the retrieved collection version
+	 * @throws IOException
+	 */
+	public String  getOclReleaseVersion(String url, String token) throws IOException {
+		String exportVersion;
+		url = removeLastUrlForwardSlashIfExist(url);
+		URL parsedUrl = new URL(url);
```
dkayiwa: A variable of name `parsedUrl` would mean a result of some sort of parsing.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
```diff
@@ -125,13 +128,73 @@ public GetMethod executeExportRequest(String url, String latestVersion) throws I
...
+	/**
+	 * This gets the desired collection release version
+	 * @param url, the subscription url
+	 * @param token, the subscription API token
+	 * @return the retrieved collection version
+	 * @throws IOException
+	 */
+	public String  getOclReleaseVersion(String url, String token) throws IOException {
+		String exportVersion;
+		url = removeLastUrlForwardSlashIfExist(url);
```
dkayiwa: Do we still need the call to removeLastUrlForwardSlashIfExist?

jecihjoy: yah definitely

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +140,42 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+
+	@Test
+	public void getExportUrl_shouldConstructCorrectExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
+		String collectionVersion = "v1.0";
+		GetMethod getMethodUrlWithoutVersion = oclClient.getExportUrl(URL_WITHOUT_VERSION, collectionVersion);
+		assertThat(getMethodUrlWithoutVersion.getURI().toString(), is(URL_WITHOUT_VERSION + "/v1.0" + "/export"));
+	}
+
+	@Test
+	public void getExportUrl_shouldConstructCorretExportUrlWhenVersionIsNotPassedToUrl() throws URIException {
```
dkayiwa: What is the reason for duplicating this test method?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -132,4 +140,42 @@ public void parseDateFromPath_shouldParseDate() throws Exception {
...
+	}
+
+	@Test
+	public void removeLastUrlForwardSlashIfExist_shouldRemoveLastForwardSlashIfItExist() {
+		String newUrl = oclClient.removeLastUrlForwardSlashIfExist(URL);
+		int numberOfForwardSlashesInUrl = StringUtils.countMatches(URL, "/");
+
+		int numberOfForwardSlashesInNewUrl = StringUtils.countMatches(newUrl, "/");
+
+		assertThat(numberOfForwardSlashesInUrl, greaterThan(numberOfForwardSlashesInNewUrl));
```
dkayiwa: If a bug is introduced in removeLastUrlForwardSlashIfExist and it removes all slashes, will this test catch it? 😊

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: Different test cases.

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
jecihjoy: I had the assertion to ensure the number of slashes is not null in my previous commit which you commented out that I should remove it

dkayiwa: Wouldn't what am asking be achieved by simply asserting that the difference in the number of slashes is one? 😊

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
dkayiwa: What is the difference between what the two methods are testing?

jecihjoy: one when the url version if provided in the subscription url and the other when the version is not provided

dkayiwa: But they both have `WhenVersionIsNotPassedToUrl`

jecihjoy: https://stackoverflow.com/questions/40278040/junit-test-if-else-case

dkayiwa: Meaning?

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
```diff
@@ -9,9 +9,7 @@
  */
 package org.openmrs.module.openconceptlab.client;
 
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.startsWith;
+import static org.hamcrest.Matchers.*;
```
dkayiwa: https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Donotusewildcardimports

api/src/test/java/org/openmrs/module/openconceptlab/client/OclClientTest.java
dkayiwa: What is the value of these two assertions?

jecihjoy: I already added this **"Wouldn't what am asking be achieved by simply asserting that the difference in the number of slashes is one?"**

dkayiwa: Let me ask it differently, is there any possibility of the second assertion passing and the one before it failing?

jecihjoy: That can't happen

Also I think we've done lot's of discussions here and I think if the implementation doesn't seem right I could move this issue to talk and let people provide their opinions on how best this feature can be implemented instead of the back and forth.

dkayiwa: Because it can't happen, as you have said, that is exactly why am saying that the first assertion is redundant. Hence it should be removed. Do you agree? 😊

PR #71, opened 2021-04-19
OCLOMRS-870: Ensure token is sent when downloading dictionary from OCL

jwnasambu:
**Ticket Worked on:**

[Unable to import a private dictionary through Subscription Module.](https://issues.openmrs.org/browse/OCLOMRS-870)

**Summary of what I did:**
Since the fetchSnapshotUpdates() function is set to “Authorization” header, but when fetching the concepts we don’t, I added the `if` block [here](https://github.com/openmrs/openmrs-module-openconceptlab/blob/e0e97a049392d88250f15419633f5d563e3a98da/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java#L119) to pass the token into this function and added the ` importService.getSubscription().getToken()` to the tests of method `executeExportRequest`. to enable import a private dictionary through Subscription Module.

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclClient.java
@@ -108,14 +108,17 @@ public class OclClient {
 
 		String collectionVersion = getOclReleaseVersion(url, token);
 
-		GetMethod exportUrlGet = executeExportRequest(url, collectionVersion);
+		GetMethod exportUrlGet = executeExportRequest(url, collectionVersion, token);
 		
 		return extractResponse(exportUrlGet);
     }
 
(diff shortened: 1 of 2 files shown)


jwnasambu:
@ibacher is the error am getting related to poor internet failing to download the required dependencies?

ibacher:
No; the error you're getting is because there's no `token` variable in the `executeExportRequest()` function. It needs to be added as a parameter to the function and then passed in from `fetchOclConcepts()`

jwnasambu:
@ibacher thanks for your response.

jwnasambu:
@ibacher kindly feel free to review this PR at your convenient time.

ibacher (approved):

ibacher:
Looks right to me. We'll need to test this out, but I'm merging this in for now. Thanks @jwnasambu!

PR #77, opened 2021-04-30
OCLOMRS-977: Update Subscription Module to support answer option order

jwnasambu:
**Issue worked on:**
[Update Subscription Module to support answer option ordering changes](https://issues.openmrs.org/browse/OCLOMRS-977)

**Summary of what I changed**

I updated the `OclMapping` class to support an` extras object` that contains a sortWeight field that contains a number. Also updated the method we use to `create or update concept answers` so that when the `extras.sortWeight` property is available we set the corresponding `ConceptAnswer's sortWeight` field to the same value.


ibacher (approved):

ibacher:
@jwnasambu Looks good! Thanks for this!

PR #82, opened 2021-09-02
OCLOMRS-1042: Mapping errors on import

moshonk:
Validation checks for referenced Concepts will now only happen if the Mapping is `Q-And-A` or `Set`

### DIFF
--- a/api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
+++ b/api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
@@ -19,9 +19,7 @@ import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -19,9 +19,7 @@
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
ibacher, line 22: Only comment on this is we don't like import `*` in OpenMRS code. Could you change to using explicit imports here?

ibacher (commented):
One small thing, otherwise, LGTM! Thanks @moshonk!

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
moshonk: Sure @ibacher, fixing that shortly. The IDE went ahead of me there :)

PR #84, opened 2021-10-02
OCLM-78 - Enable importing single concept exports from OCL

mseaton:
This adds the ability to import a single concept export from OCL into an OpenMRS instance via the existing import page.  See demo below of the code within this PR in action:

ocl-import

What this PR does:

* Removes the "zip" file restriction, and supports uploading _either_ a zip file _or_ a json file, both for the existing Collection import, and for the new single-concept import
* Adds a new subdirectory to the the existing .OpenMRS/ocl directory for .OpenMRS/ocl/imports.  
** Copies any uploaded file into this "imports" directory once it has an import attempt, with a timestamp differentiator
** Changes the existing logic that set the import URL to the .OpenMRS/ocl/configuration/loadAtStartup for file imports, and instead set the import URL to the path to the saved file within the imports directory
* Adding the necessary logic to the Importer class to support:
** Importing either zip or json
** Importing a single concept in addition to a collection
* Changed Saver.saveMapping to look existing Concepts up by mapping code rather than by uuid (this was necessary as the concept I'm importing had answers that exist in my dictionary with CIEL mappings already on them, but without a CIEL uuid).  Looking up by mapping allowed me to import this concept and associate the correct answers without re-importing those answers or having to change my existing uuids.  Note: this likely will solve the same problem that Ellen reported in OCLOMRS-1008
* Updated the UI to allow for uploading json as well as zip files, and for uploading single-concept exports as well as collection exports.


api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
```diff
@@ -113,82 +175,179 @@ public void setZipFile(ZipFile zipFile) {
...
+									oclMapping.setFromSourceUrl(prependBaseUrl(baseUrl, oclMapping.getFromSourceUrl()));
+									oclMapping.setToConceptUrl(prependBaseUrl(baseUrl, oclMapping.getToConceptUrl()));
+									// We do not import Q-AND-A mappings for which the imported concept is an answer
+									boolean isQuestionAnswer = "Q-AND-A".equals(oclMapping.getMapType());
+									if (isQuestionAnswer && oclMapping.getToConceptUrl().equals(oclConcept.getUrl())) {
+										log.debug("Skipping Q-AND-A mapping for which the imported concept is a mapped answer");
+									}
+									else if (oclMapping.getToSourceName().equals(oclConcept.getSource()) && oclMapping.getToConceptCode().equals(oclConcept.getId())) {
+										log.debug("Skipping SAME-AS Mapping for OCL ID as this is added in the ImportTask, and would result in duplicate mappings");
+									}
```
ibacher, line 233: I'll be honest: this probably a guard we should have in the `ImportTask` rather than here.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
```diff
@@ -113,82 +175,179 @@ public void setZipFile(ZipFile zipFile) {
...
+									OclMapping oclMapping = mappingsParser.readValueAs(OclMapping.class);
+									oclMapping.setUrl(prependBaseUrl(baseUrl, oclMapping.getUrl()));
+									oclMapping.setFromConceptUrl(prependBaseUrl(baseUrl, oclMapping.getFromConceptUrl()));
+									oclMapping.setFromSourceUrl(prependBaseUrl(baseUrl, oclMapping.getFromSourceUrl()));
+									oclMapping.setToConceptUrl(prependBaseUrl(baseUrl, oclMapping.getToConceptUrl()));
+									// We do not import Q-AND-A mappings for which the imported concept is an answer
+									boolean isQuestionAnswer = "Q-AND-A".equals(oclMapping.getMapType());
+									if (isQuestionAnswer && oclMapping.getToConceptUrl().equals(oclConcept.getUrl())) {
+										log.debug("Skipping Q-AND-A mapping for which the imported concept is a mapped answer");
+									}
```
ibacher, line 230: Is this a common occurrence? Also, isn't this something that should be fixed outside of this module, i.e., at the concept source itself?

owa/app/js/home/home.controller.js
```diff
@@ -35,15 +35,15 @@ class HomeController {
     }
     
     function isFileCorrect(fileName) {
-      return fileName.endsWith(".zip");
+      return fileName.endsWith(".zip") || fileName.endsWith(".json")
     }
 
-    function uploadZip(file) {
+    function uploadZip(file, importType) {
```
ibacher, line 41: SHould probably rename this to `uploadFile()`, right?

ibacher (commented):
Generally I'm in favour of these changes, especially looking up concepts by mappings before assuming they don't exist, but I think we need some additional discussion around this point:

> instead set the import URL to the path to the saved file within the imports directory

The problem here is that this seems to break the "happy path" flow for OCL import. As I understand things, the idea with file import wasn't "this is a quick way to get concepts into your dictionary", but rather "our site can't connect to the OCL server; here's a way we can load the dictionary offline."

In other words, the subscription URL used would be the same regardless of whether the concept came from an export from OCL or was uploaded via a zip file. This change breaks that flow as concepts imported via a zip file will no longer have the same "identity" as concepts imported via a download from OCL.

Finally, I think there's a slightly more philosophical question: does this module exist to get content from OCL into OpenMRS or does it exist to allow OCL to be used to manage OpenMRS dictionaries, because those are slightly different outlooks that lead to somewhat different conclusions. I.e., if the purpose of this module is just to get content from OCL into OpenMRS, then this is a logical bit of functionality to have. However, if the purpose is to have collection in OCL and be able to say "this collection reflects what's on my server", this functionality kind of breaks that since what is actually on the server depends on the order in which things were processed (e.g., if there's a conflict between what was loaded locally from a file and what OCL says).

mseaton:
> This change breaks that flow as concepts imported via a zip file will no longer have the same "identity" as concepts imported via a download from OCL

I admit to not really understanding yet the nature of the importance of this import URL, and it was not clear to me that this had anything to do with the "identity" of a concept.  It isn't clear to me that the current implementation - which just sets the import URL to the location specified in the global property "ocl/configuration/loadAtStartup" - was actually done meaningfully, as this a) doesn't point to a specific file, and b) any file placed in there is immediately deleted once it is processed at startup.

> Finally, I think there's a slightly more philosophical question: does this module exist to get content from OCL into OpenMRS or does it exist to allow OCL to be used to manage OpenMRS dictionaries,

This is certainly a fair point.  I don't honestly know.  I approached this as follows:

1. I wanted to get a single concept onto my server from OCL.  I wanted it to have the same UUID as the one defined on OCL.  There is no way to do this in OpenMRS other than via direct SQL (you can't set a concept's uuid via the create concept UI).  And I was able to download a structured JSON file from OCL that had everything in it that I needed, so this felt like a good place to start.

2. The "openconceptlab" module already has almost everything I needed in place (or at least I thought it did, it turned out that it didn't entirely).  In that - it already had the model objects and parsers for reading in an OCL JSON export, and it already had the importers in place to create concepts/names/mappings from this Object model.  So, it seemed like it made sense to build on this and not re-invent the wheel in yet another module

Personally I'd like to see this module be about how to usefully and practically manage (or simply find shared) Concepts defined in OCL and import those into one's dictionary.  Subscription is one possible approach to this, but to be honest, in terms of getting comfortable adopting OCL, I'm much more comfortable starting out with a single concept, whose impact I can clearly understand and test on my dictionary, than starting with a subscription.

If we want to make this purely about subscription, that's fine, but we might find it useful to develop a small shared library (i.e. like the hapi APIs) that can be shared and used for reading OCL export files and converting these to OpenMRS concepts.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
```diff
@@ -97,8 +97,70 @@ public void setSaver(Saver persister) {
...
+				importService.updateSubscriptionUrl(anImport, subscription.getUrl());
+				in = new CountingInputStream(oclResponse.getContentStream());
+				totalBytesToProcess = oclResponse.getContentLength();
+			}
+		}
+	}
+
+	private void closeInputStream() {
+		IOUtils.closeQuietly(in);
+	}
```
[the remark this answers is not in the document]
mseaton: I'll need to look back at the code but there was a reason I needed to do it this way.  I think it was because there are 2 paths here - one in which an Input Stream is coming from a File, and another in which an Input Stream is coming from the subscription.  If we can make this work with a try-with-resources, I'm all for it, but we'd need to look at why this is like this.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
mseaton: Yes, it is common - anytime you are exporting a concept that is the answer to a question.  Look at the export for "TRUE" in OCL - it has 710 "associations", most of which are Q-AND-A mappings for which it is an answer:
https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/1065/

The assumption / decision I made was to not worry about these associations when exporting a concept that is an answer, only worry about those when exporting a question that has this answer.  Make sense?

owa/app/js/home/home.controller.js
mseaton: Yes, good catch

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
ibacher: Right... the custom code was added to work around the fact that the import of CIEL wasn't including those SAME-AS maps (and to deal with the fact that dictionaries don't necessarily have the SAME-AS mapping to their concept id.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Importer.java
ibacher: I understand that associations like this are common, but they aren't generally included in the file, depending on how it's generated. E.g., when I download that concept as JSON from the OCL Term Browser, it actually has no mappings in it at all.

[Attachment](https://github.com/openmrs/openmrs-module-openconceptlab/files/7367590/concept-1065.json.txt)

If I download the concept from the API, I can either get the concept with no mappings at:

https://api.openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1065/

Or with mappings where the concept is the `from` concept:

https://api.openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1065/?includeMappings=true

Or with mappings where the concept is the `to` concept:

https://api.openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1065/?includeInverseMappings=true

I can also combine those filter. The point was that, in general, I wouldn't expect someone to end up with a bunch of mappings where the concept is the to concept and if we do we need to guard against it more generally.

mseaton: That's cool that there is that flexibility in the API.  Unfortunately, that same flexibility isn't in the term browser, and the term browser that I see does contain all of those mappings.  This is what I was using to download a Concept.  My experience with OCL is that there is a lot of inconsistency in each of the various pages you might go to in terms of what is included in an export file.  For a Concept like this you get either JSON or ZIP.  For other Collections or Dictionaries, you sometimes get JSON, sometimes CSV, sometimes ZIP formats, and these formats are not necessarily consistent with each other (as you say, sometimes they contain mappings and other details, sometimes they do not).

ibacher: Ah... that's fair. Actually this would be a good thing to put more centrally in mapping exclusions because I suppose the export format could change whether intentionally or accidentally to include inverse mappings.

PR #86, opened 2021-11-09
OCLOMRS-1067: Save sort weights when adding or updating concept answers

mseaton:
…ers and concept set members.


mogoodrich (approved):
Defer to @ibacher  because he's more of an expert on this, but looks good to me!

ibacher (commented):
LGTM

PR #88, opened 2021-11-15
OCLOMRS-1008: Use existing mapped concepts instead of creating new ones

ibacher:
…e concept exists

This includes one of the key features from #84 but in a slightly different way (instead of using `ConceptService#getConceptByMapping()` this includes a custom implementation that depends on the SAME-AS mapping).

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
@@ -12,12 +12,12 @@ public interface OclConceptService {
 	 * which returns concepts that are marked as the "SAME-AS" an existing concept
 	 *
 	 * @param source the name of the source to check
-	 * @param conceptId the identifier of the concept
+	 * @param code the identifier of the concept
 	 * @return the concept that is semantically the same as the requested concept or null if
 	 *  one couldn't be found.
 	 */
 	@Transactional(readOnly = true)
 	@Authorized(PrivilegeConstants.VIEW_CONCEPTS)
-	Concept getDuplicateConceptByMapping(String conceptId, String source);
+	Concept getConceptWithSameAsMapping(String code, String source);
 	
 }
--- a/api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
@@ -1,0 +1,23 @@
+package org.openmrs.module.openconceptlab;
+
+import org.openmrs.Concept;
+import org.openmrs.annotation.Authorized;
+import org.openmrs.util.PrivilegeConstants;
+import org.springframework.transaction.annotation.Transactional;
+
+public interface OclConceptService {
+	
+	/**
+	 * A version of {@link org.openmrs.api.ConceptService#getConceptByMapping(String, String)}
+	 * which returns concepts that are marked as the "SAME-AS" an existing concept
+	 *
+	 * @param source the name of the source to check
+	 * @param conceptId the identifier of the concept
+	 * @return the concept that is semantically the same as the requested concept or null if
+	 *  one couldn't be found.
+	 */
+	@Transactional(readOnly = true)
+	@Authorized(PrivilegeConstants.VIEW_CONCEPTS)
+	Concept getDuplicateConceptByMapping(String conceptId, String source);
+	
+}
--- a/api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
@@ -91,8 +91,13 @@ public class Saver {
 		
 		// if a mapping for this concept already exists and it is not
 		if (item == null && oclConcept.getSource() != null && oclConcept.getId() != null) {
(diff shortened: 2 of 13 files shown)


ibacher:
I'll be honest, I don't understand the test failure. I'll look into it tomorrow.

api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
```diff
@@ -0,0 +1,23 @@
...
+	 * which returns concepts that are marked as the "SAME-AS" an existing concept
+	 *
+	 * @param source the name of the source to check
+	 * @param conceptId the identifier of the concept
+	 * @return the concept that is semantically the same as the requested concept or null if
+	 *  one couldn't be found.
+	 */
+	@Transactional(readOnly = true)
+	@Authorized(PrivilegeConstants.VIEW_CONCEPTS)
+	Concept getDuplicateConceptByMapping(String conceptId, String source);
```
mseaton, line 21: I find this method name confusing.  Can we just call it "getConceptWithSameAsMapping"?  And name "conceptId" -> "code"?

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -26,24 +13,49 @@
...
-	
+
+	<bean parent="serviceContext">
+		<property name="moduleService">
+			<list>
+				<value>org.openmrs.module.openconceptlab.OclConceptService</value>
+				<ref bean="openconceptlab.conceptService"/>
+			</list>
+		</property>
+	</bean>
```
mseaton, line 57: Do we need to list this twice, or can we list it once and have the two services in the list?

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -88,6 +88,12 @@ public Item saveConcept(final CacheService cacheService, final Import anImport,
 		if (item != null && item.getVersionUrl().equals(oclConcept.getVersionUrl())) {
 			return new Item(thisImport, oclConcept, ItemState.UP_TO_DATE);
 		}
+		
+		if (oclConcept.getSource() != null && oclConcept.getId() != null) {
+			if (cacheService.getConceptByMapping(oclConcept.getSource(), oclConcept.getId()) != null) {
+				return new Item(thisImport, oclConcept, ItemState.UP_TO_DATE);
+			}
+		}
```
mseaton, line 96: Can you explain this block?  I may need to see it better in context, but might this prevent updates from happening?

mseaton (commented):
Few comments @ibacher - if there are build failures I can try to build locally and see what I find...

mseaton:
> I'll be honest, I don't understand the compilation failure. I'll look into it tomorrow

I had a weird experience the other day with a build failure I couldn't explain.  I ended up fixing it by removing the Maven cache step from the github action.  It had ended up with a bad artifact in the cache and was causing it to fail...

api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
ibacher: Well, the name arose because I was planning on eventually including Burke's "CODE" mapping if / when that comes about, but yeah, I suppose that's more accurate.

api/src/main/resources/moduleApplicationContext.xml
ibacher: This is probably something we should fix at some point, but [the `setModuleServiceMethod()`](https://github.com/openmrs/openmrs-core/blob/011d738ee5f1a325d38c2cac588fe7446de7885a/api/src/main/java/org/openmrs/api/context/ServiceContext.java#L739) only processes the first two arguments, so it's necessary to invoke this twice.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
ibacher: This needs an extra guard, but the idea is this: If I have a concept in my dictionary marked SAME-AS CIEL:1066, then we shouldn't load the concept CIEL:1066 into the concept dictionary. The extra guard is to make sure that that item isn't the same as the item we're importing (based on the OCL URL).

Originally, I had this written to return an `ItemState.DUPLICATE`, but I wasn't sure it was worth the effort to add a new `ItemState` for this.

mseaton:
> I had a weird experience the other day with a build failure I couldn't explain. I ended up fixing it by removing the Maven cache step from the github action. It had ended up with a bad artifact in the cache and was causing it to fail...

Looks like you just tried this @ibacher and it didn't fix it.  Weird.  I'm running locally on Ubuntu 20 with Maven 3.6.3 and OpenJDK 1.8.0_292 and things work.  Looks like GA is using a JDK from Azul Systems, version 1.8.0_212 and Maven 3.8.3.  I suppose that might be where the issue lies?

mseaton (approved):
LGTM, assuming we can get it to pass the build in github :)

ibacher:
Yeah... it works for me on AdoptOpenJDK 1.8.0_292 and Maven 3.8.1 locally. I've tried adjusting the build to use a simpler workflow and AdoptOpenJDK so it's as close to my environment as possible.

PR #89, opened 2021-11-17
Oclomrs 1076

mseaton:


api/pom.xml
```diff
@@ -12,60 +12,6 @@
...
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-compress</artifactId>
-            <version>1.21</version>
-        </dependency>
-
-    </dependencies>
-
```
mseaton, line 68: When I added the 2.2 profile, it took some trial and error to get things to pass.  To make things a little easier to figure out, I moved all of the shared api/omod dependencies into the top-level pom, and all of the omod-specific dependencies into the omod pom, and got rid of dependencyManagement.

api/src/main/java/org/openmrs/module/openconceptlab/OclConceptService.java
```diff
@@ -17,7 +15,6 @@
 	 *  one couldn't be found.
 	 */
 	@Transactional(readOnly = true)
-	@Authorized(PrivilegeConstants.VIEW_CONCEPTS)
```
mseaton, line 20: @ibacher I had to remove this Authorized annotation, as this privilege is no longer in 2.x.  It's unfortunate we didn't just keep the constant as-is and change the value of it.

api/src/main/java/org/openmrs/module/openconceptlab/Utils.java
```diff
@@ -177,4 +181,31 @@ public static UUID version5Uuid(String name) {
...
+		} catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+			try {
+				Method getAllowDecimal = numeric.getClass().getDeclaredMethod("getAllowDecimal");
+				return (Boolean) getAllowDecimal.invoke(numeric);
+			} catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e1) {
+				throw new ImportException(e1);
+			}
+		}
+	}
 }
```
mseaton, line 211: I need a "getAllowDecimal" method for some unit tests to pass in both pre-2x and post-2x, so I moved these two methods here so they would be together for easier reference and reuse.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -665,13 +656,29 @@ private void updateOrAddNamesFromOcl(Concept concept, OclConcept oclConcept) {
...
 				}
 				name.setConceptNameType(oclNameType);
-				name.setLocalePreferred(oclName.isLocalePreferred());
+
+				// If this concept does not yet have any names preferred in the given locale, make it preferred
+				// This is needed in order to match expected state in the Concept.addName method
+				boolean preferred = oclName.isLocalePreferred();
+				if (!preferred) {
+					preferred = !hasExistingPreferredNameInExactLocale(concept, oclName.getLocale());
+				}
```
mseaton, line 665: This is ultimately all that this PR is really about - this is the bug fix I am introducing.  Everything else in this PR is about demonstrating the issue and the fix through unit tests (which necessitate the 2.x profile, etc)

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -665,13 +656,29 @@ private void updateOrAddNamesFromOcl(Concept concept, OclConcept oclConcept) {
...
 
+	private boolean hasExistingPreferredNameInExactLocale(Concept concept, Locale locale) {
+		for (ConceptName cn : concept.getNames()) {
+			if (BooleanUtils.isTrue(cn.getLocalePreferred()) && cn.getLocale().equals(locale)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
```
mseaton, line 681: The may be a bit repetitive of what is in the core Concept.getNames(Locale) method, but one of the reasons why this bug emerged is that the implementation of these methods in core has changed over time and isn't always clear when the Locale is being matched exactly and when it falls back to Language.  I didn't want to leave this to chance, so just am implementing a version here that contains exactly the logic we need.

api/src/main/java/org/openmrs/module/openconceptlab/scheduler/UpdateScheduler.java
```diff
@@ -26,7 +26,7 @@
 	
 	ThreadPoolTaskScheduler scheduler;
 	
-	ScheduledFuture<Importer> scheduledUpdate;
+	ScheduledFuture<?> scheduledUpdate;
```
mseaton, line 30: This was needed for 2.x compatibility (for Java 1.8 I think)

api/src/test/java/org/openmrs/module/openconceptlab/ImportServiceTest.java
```diff
@@ -302,7 +303,7 @@ public void update_shouldUpdateReleaseVersion() throws Exception {
     }
 
 	@Ignore("This test passes locally, but fails on Travis-CI")
-	@NotTransactional
+	@Transactional(propagation = Propagation.NEVER)
```
mseaton, line 306: All of the tests that use this annotation are @Ignored so I guess I don't know if this is the right fix.  I'm just noticing that now.  I guess we should remove the @Ignore annotation and see what happens with these.  Maybe Github Actions will pass where Travis-CI did not...

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -172,6 +198,68 @@ private Name newFourthName() {
 		return fourthName;
 	}
 
+	/**
```
mseaton, line 201: This is the unit test that demonstrates the bug for this ticket.

omod/pom.xml
```diff
@@ -12,6 +12,26 @@
...
+                <dependency>
+                    <groupId>javax.servlet</groupId>
+                    <artifactId>javax.servlet-api</artifactId>
+                    <version>4.0.1</version>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
+
```
mseaton, line 34: If / when we merge this, we should add a stage to Bamboo to build and test the 2.2 profile

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/search/ImportByRunningImportSearchHandler.java
```diff
@@ -35,7 +35,7 @@ public void setImporter(Importer importer) {
     private static final String RUNNING_IMPORT = "runningImport";
 
     private final SearchConfig searchConfig = new SearchConfig("default", RestConstants.VERSION_1 + OpenConceptLabRestController.OPEN_CONCEPT_LAB_REST_NAMESPACE + "/import",
-            Arrays.asList("1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*"),
+            Arrays.asList("1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*"),
```
mseaton, line 38: These things are super annoying.  Presumably we should also add in 2.3.x, 2.4.x,... to this and the one below too.  I just added the 2.2.* to get the unit tests to pass in the 2.2 profile, but this will likely cause someone else headaches in the future if we don't update this.

pom.xml
```diff
@@ -55,100 +55,74 @@
 
     <properties>
         <openmrsVersion>1.9.11</openmrsVersion>
-        <!-- Module dependencies -->
-        <webservices.restVersion>2.17</webservices.restVersion>
+        <legacyuiVersion>1.9.0</legacyuiVersion>
+        <webservices.restVersion>2.29.0</webservices.restVersion>
```
mseaton, line 59: Bumped up the rest webservices module to the latest release version

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -172,6 +198,68 @@ private Name newFourthName() {
 		return fourthName;
 	}
 
+	/**
+	 * @see Saver#saveConcept(CacheService, Import, OclConcept)
+	 * @verifies add new names to concept
+	 */
+	@Test
+	public void importConcept_shouldAddNewPreferredNamesInDifferentLocales() throws Exception {
```
ibacher, line 206: My only nit with this is it would be better to name it `saveConcept()`, since that's the method being tested. I've been doing that with newer tests, but haven't (yet) migrated existing tests.

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/search/ImportByRunningImportSearchHandler.java
ibacher: Actually, this explains [OCLM-79](https://issues.openmrs.org/browse/OCLM-79).

I'm inclined to replace this with `Collections.singletonList("1.8.* - 2.*")` and similarly elsewhere.

ibacher (approved):
One small nit, but this LGTM. Thanks @mseaton!

omod/pom.xml
ibacher: We should do the same for Github Actions, so we can catch it on PRs to...

omod/src/main/java/org/openmrs/module/openconceptlab/web/rest/search/ImportByRunningImportSearchHandler.java
mseaton: Done.  Replaced in these 2 classes.  Thanks for pointing out that you can do that.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
mogoodrich: Not quite sure I understand what the issue was and why this was the fix?

PR #90, opened 2021-11-17
Add back maven caching to github actions.  Remove node from repositor…

mseaton:
…y and add to gitignore.


mseaton:
@ibacher - can I get your opinion on these.  Let's get the cache back in the github action build now that we know that wasn't causing any issues.  And let's get rid of this committed version of node.

.github/workflows/maven.yml
```diff
@@ -27,6 +27,13 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: ${{ matrix.java-version }}
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
ibacher, line 36: Based on [these docs](https://github.com/marketplace/actions/setup-java-jdk#caching-packages-dependencies), I think this can just be done with this:

```suggestion
          java-version: ${{ matrix.java-version }}
          cache: 'maven'
```

ibacher (approved):

ibacher:
Apart from the build failure, I think this is great!

.github/workflows/maven.yml
mseaton: Sorry @ibacher - I didn't see this suggestion when I merged this.  Feel free to add this in / tweak, but things just passed :)

PR #91, opened 2021-11-17
OCLOMRS-1077:  Saver should set the "set" property on Concept based o…

mseaton:
…n the "is_set" value in extras.

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
@@ -379,6 +379,9 @@ public class OclConcept {
 	@JsonIgnoreProperties(ignoreUnknown = true)
 	public static class Extras {
 
+		@JsonProperty("is_set")
+		private Integer isSet;
+
 		@JsonProperty("hi_absolute")
 		private Double hiAbsolute;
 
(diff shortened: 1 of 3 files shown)


ibacher:
@mseaton Do we need this feature? We already have things setup so that as soon a concept set member is found, the concept should be marked as a concept set [see here](https://github.com/openmrs/openmrs-module-openconceptlab/blob/e19850c708461bae3c2bc9bc1b59dd9f75eea783/api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java#L395). Is that not working or do we need to handle empty concept sets?

mseaton:
It's a good question @ibacher .  Basically this ticket came about as I tested this general process:

1. Create "source DB" of concepts pre-OCL.
2. Export these concepts and import them into OCL, and then export them from OCL into an export zip
3. Read this export zip into an empty OMRS instance using the subscription module to create "target DB".
4. Compare source DB vs. target DB

In this analysis I found that the is_set property on concept was inconsistent in the before / after.

I'm making the assumption that if is_set is explicitly set, and this is explicitly added as an extra in OCL, then we want to honor that above any assumption we make around actual set members (eg. there may be cases where concepts are created to be sets but don't - yet - have any members).

ibacher:
> I'm making the assumption that if is_set is explicitly set, and this is explicitly added as an extra in OCL, then we want to honor that above any assumption we make around actual set members

Yes, that's true, I didn't realise that `is_set` was being set on the extras field at all. It would be nicer if it were true / false rather than 1 / 0, but in any case, this LGTM.

mogoodrich (approved):
LGTM, thanks @mseaton

mseaton:
> It would be nicer if it were true / false rather than 1 / 0

Agree.  If OCL were to change this, we can evolve this to match.

PR #92, opened 2021-11-17
OCLOMRS-1074 - Need to import ConceptComplex where appropriate

mseaton:
This will depend on exactly how @bmamlin adds in the "handler" to the Extras in OCL, but this PR is written with the assumption that this will be called "handler".

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
@@ -9,11 +9,6 @@
  */
 package org.openmrs.module.openconceptlab.client;
 
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-
 import org.apache.commons.lang3.builder.EqualsBuilder;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
```diff
@@ -405,6 +405,9 @@ public String toString() {
 		@JsonProperty("allow_decimal")
 		private Boolean allowDecimal;
 
+		@JsonProperty("handler")
+		private String handler;
```
mogoodrich, line 409: I guess we might want to consider calling this "complexHandler"?  But not a strong preference, and I guess it depends on what Burke is calling this.

mogoodrich (approved):
LGTM!

api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
mseaton: @bmamlin called this "handler", so keeping this as-is

PR #104, opened 2024-05-01
OCLOMRS-1111 OCL should stop and fail any in progress imports at startup

mseaton:
See:  https://openmrs.atlassian.net/browse/OCLOMRS-1111

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
@@ -103,14 +103,14 @@ public class OpenConceptLabActivator extends BaseModuleActivator implements Daem
 	 * In this situation, we want to ensure these imports are marked as failed at the next startup, so that
 	 * subsequent import attempts can succeed without errors that imports are already in progress
 	 */
-	protected void markInProgressImportsAsFailed() {
+	private void markInProgressImportsAsFailed() {
 		ImportService importService = Context.getService(ImportService.class);
 		List<Import> inProgressImports = importService.getInProgressImports();
 		if (!inProgressImports.isEmpty()) {
 			log.warn("Found " + inProgressImports.size() + " in progress imports at startup");
 			for (Import inProgressImport : inProgressImports) {
 				log.warn("Updating Import#" + inProgressImport.getImportId() + " as failed");
-				importService.failImport(inProgressImport, "System restarted during import");
+				importService.failImport(inProgressImport, "System interruption during import");
 				importService.stopImport(inProgressImport);
 			}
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -93,6 +97,25 @@ public void contextRefreshed() {
 		log.info("Open Concept Lab Module refreshed");
 	}
 
+	/**
+	 * If OpenMRS is unexpectedly shutdown while an OCL Import is in progress, the import record in the database
+	 * can be left in an "in progress" state, even though the import is no longer actually running.
+	 * In this situation, we want to ensure these imports are marked as failed at the next startup, so that
+	 * subsequent import attempts can succeed without errors that imports are already in progress
+	 */
+	protected void markInProgressImportsAsFailed() {
```
dkayiwa, line 106: Could there be a reason why this method is not private?

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
```diff
@@ -93,6 +97,25 @@ public void contextRefreshed() {
...
+		ImportService importService = Context.getService(ImportService.class);
+		List<Import> inProgressImports = importService.getInProgressImports();
+		if (!inProgressImports.isEmpty()) {
+			log.warn("Found " + inProgressImports.size() + " in progress imports at startup");
+			for (Import inProgressImport : inProgressImports) {
+				log.warn("Updating Import#" + inProgressImport.getImportId() + " as failed");
+				importService.failImport(inProgressImport, "System restarted during import");
```
mogoodrich, line 113: super nit, but this also could happen if the context is refreshed (ie someone manually loads a module via the old UI) while an import is rerunning.  Just wondering if the message should be "Context refreshed and/or system restarted during import."

mogoodrich (approved):
LGTM, one comment, feel free to ignore, especially if I'm wrong... :)

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
mseaton: OK, changed it to private @dkayiwa

api/src/main/java/org/openmrs/module/openconceptlab/OpenConceptLabActivator.java
mseaton: @mogoodrich thanks I updated the message.

dkayiwa (approved):

PR #109, opened 2024-10-21
OCLOMRS-1116: Retired Concept Answers and Concept Set Members get re-…

mogoodrich:
…added to concepts

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
@@ -43,6 +43,7 @@ import org.openmrs.module.openconceptlab.client.OclMapping.MapType;
 import org.openmrs.util.OpenmrsUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.config.BeanPostProcessor;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -566,7 +567,7 @@ public class Saver {
 			}
 		}
 
-		if (!found) {
+		if (!found && BooleanUtils.isNotTrue(oclMapping.getRetired())) {
 			ConceptSet conceptSet = new ConceptSet();
 			conceptSet.setConceptSet(set);
 			conceptSet.setConcept(member);
@@ -574,6 +575,8 @@ public class Saver {
 			conceptSet.setSortWeight(getSortWeightForMapping(oclMapping, 1.0));
 			set.getConceptSets().add(conceptSet);
 			item = new Item(update, oclMapping, ItemState.ADDED);
+		} else {
+			item = new Item(update, oclMapping, ItemState.UP_TO_DATE);
 		}
 
 		return item;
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -1221,6 +1274,31 @@ public void importMapping_shouldRemoveConceptMapping() throws Exception {
 		assertThat(BooleanUtils.isTrue(term.getRetired()), is(false));
 	}
 
+	@Test
```
mogoodrich, line 1277: This test passes even without this change, I just wrote it to verify that a similar problem wasn't happening with concept mappings.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -566,14 +567,16 @@ Item updateOrAddSetMembersFromOcl(Import update, OclMapping oclMapping, Concept
...
+		if (!found && BooleanUtils.isNotTrue(oclMapping.getRetired())) {
 			ConceptSet conceptSet = new ConceptSet();
 			conceptSet.setConceptSet(set);
 			conceptSet.setConcept(member);
 			conceptSet.setUuid(oclMapping.getExternalId());
 			conceptSet.setSortWeight(getSortWeightForMapping(oclMapping, 1.0));
 			set.getConceptSets().add(conceptSet);
 			item = new Item(update, oclMapping, ItemState.ADDED);
+		} else {
+			item = new Item(update, oclMapping, ItemState.UP_TO_DATE);
```
mogoodrich, line 579: This seems like what we want to report here, but interested in others opinins.

mogoodrich:
Whoops, converting this to draft, on other thing to fix!

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -603,13 +608,16 @@ private Item updateOrAddAnswersFromOcl(Import update, OclMapping oclMapping, Con
...
-		}
+		if (!found)
+			if(BooleanUtils.isNotTrue(oclMapping.getRetired())) {
+				ConceptAnswer conceptAnswer = new ConceptAnswer(answer);
+				conceptAnswer.setUuid(oclMapping.getExternalId());
+				conceptAnswer.setSortWeight(getSortWeightForMapping(oclMapping, null));
+				question.addAnswer(conceptAnswer);
+				item = new Item(update, oclMapping, ItemState.ADDED);
+			} else {
+				item = new Item(update, oclMapping, ItemState.UP_TO_DATE);
```
mogoodrich, line 619: This seems like the correct state to report here, but interested in others thoughts.

mogoodrich (commented):
Okay, updated, now ready for review again!

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
ibacher: Seems right. `UP_TO_DATE` is intended to mark exactly this, I think.

mseaton (approved):

PR #115, opened 2026-02-02
OCLOMRS-1120: Switch to a non-thread-pooled implementation processing imports

ibacher:
In the RefApp we've been running into issues with OCL imports that seem related to the relatively complex multi-threaded setup used (essentially, all concepts and mappings were chunked into chunks of 128 and each chunk was processed in it's own thread). This seems to cause issues in some cases where either concepts are not imported correctly or (more likely) that Hibernate was incorrectly reporting that concepts were not imported due to cache flushing issues, etc. This PR switches the module to running in a simpler, sequential fashion.

On my machine loading the RefApp (which has several different imports that need to run), this runs in 335 seconds vs 255 seconds with the parallel version. Not sure if that performance is acceptable, although it is more correct.

Note that the imports are still run in background threads, but only a single background thread per import.


ibacher:
@mseaton @dkayiwa I'm tagging you as reviewers to get your input on whether or not this is a good idea.

dkayiwa:
@ibacher if we required Java 21, do you think virtual threads would lead to any performance improvements?

mseaton:
> @mseaton @dkayiwa I'm tagging you as reviewers to get your input on whether or not this is a good idea.

I am not particularly worried about an extra 1-2 minutes of load time as this only happens when there is a new package to install, which will only happen for a new release, not on every restart / deploy.  Obviously we want to keep these numbers down as much as possible, but I'd value stability over speed for this kind of operation

ibacher:
> if we required Java 21, do you think virtual threads would lead to any performance improvements?

Probably not very much. Running the OCL imports is heavily IO bound, so the CPU is substantially under-utilized. The reason any kind of threading (virtual or normal) might make sense here is because arguably the chunks the importer works through _should_ be independent (the OCL export is broken up into "concepts" and "mappings" sections and the importer basically reads each of those in turn).

It is possible that just more aggressively flushing and clearing the session would yield the same "correctness" results... I haven't explored that and probably should.

dkayiwa (approved):

PR #119, opened 2026-03-02
Optimize large OCL zip file import performance

dkayiwa:
## Summary

Optimizes the OCL zip file import pipeline for large files (e.g., DiagnosesStarterKit with ~5000 concepts and ~10000 mappings). The main bottleneck was excessive per-item database queries during import.

### 1. Item URL Cache in CacheService (biggest win)
- Added in-memory `HashMap` cache for `getLastSuccessfulItemByUrl()` results (including null/not-found)
- Cache persists across `clearCache()` calls since items are used for metadata only (uuid, versionUrl, state)
- Items created during concept phase are cached for instant lookup during mapping phase
- **Eliminates ~35,000+ redundant DB queries** for a typical large import

### 2. Skip DB Item Lookups for First-Time Imports
- Detects first-ever import (`getImportsInOrder` returns ≤1 result)
- Sets `skipDbItemLookups` flag on CacheService to skip `getLastSuccessfulItemByUrl()` DB queries
- For first imports, no previous items exist so DB queries always return null
- Flag is scoped only to Item URL lookups — ConceptMap and other entity lookups always query the DB on cache miss, since those entities may exist from non-OCL sources
- **Eliminates ~15,000+ guaranteed-null DB queries** on initial import

### 3. ConceptMap Cache in CacheService
- Routes `getConceptMapByUuid()` through CacheService instead of direct ImportService call
- Caches results to avoid repeated lookups for the same UUID within and across batches
- **Reduces redundant ConceptMap DB queries** during mapping phase

### 4. Validation Type Determined Once Per Import
- Determines `ValidationType` once at the start of `processInput()` and passes it through to `saveConcept()`
- Defaults to `FULL`; only overridden if a subscription exists and has an explicit `validationType` set
- The 3-arg `saveConcept()` overload (used by tests/other callers) retains the original per-call `getSubscription()` fallback for backward compatibility
- **Eliminates ~5,000 redundant `getSubscription()` calls** during the import path

### 5. Subscription Lookup Consolidation
- Looks up subscription once at the start of `processInput()` instead of 3 separate calls
- Minor optimization but eliminates redundant DB queries per import

### 6. Increased Batch Size (256 → 512)
- Reduces the number of flush/clear/reload cycles during import
- Fewer cache rebuilds and session management overhead
- Conservative increase (not 1024) to balance performance with Hibernate session memory usage, since each concept carries names, descriptions, mappings, etc. and OpenMRS deployments vary widely in available memory

### 7. Reduced Logging Overhead
- Changed per-item `log.info()` to `log.debug()` for concept/mapping import messages
- Keeps error logging at ERROR level
- **Eliminates ~15,000 log entries** with full object `toString()` calls

## Estimated Performance Improvement

For a large zip file with ~5000 concepts and ~10000 mappings:
- **~60-70% reduction in total database queries** (from ~100,000+ to ~30,000-40,000)
- **~45-55% total wall-clock time improvement** (DB queries are the dominant cost)

## Testing

All 87 existing tests pass (0 failures, 0 errors, 8 skipped - pre-existing).

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/CacheService.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/CacheService.java
@@ -21,9 +21,21 @@ import org.openmrs.api.ConceptService;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
+/**
+ * In-memory cache layer for the OCL import pipeline. A new instance is created per import run
+ * in Importer.processInput() and is garbage collected when the import completes.
+ * <p>
+ * Most entity caches (concepts, conceptMaps, etc.) are cleared on each flush/clear cycle via
+ * {@link #clearCache()}. The item URL caches ({@code itemsByUrl}, {@code checkedItemUrls}) grow
+ * monotonically for the lifetime of the import since they must persist across batches and phases.
+ * For typical imports (~15K items), this is a few MB. For very large imports (hundreds of thousands
+ * of items), memory usage should be monitored.
+ */
 public class CacheService {
 
 	ConceptService conceptService;
@@ -58,6 +70,11 @@ public class CacheService {
 
 	// Cache for ConceptReferenceTerms keyed by "sourceId:code"
 	Map<String, ConceptReferenceTerm> conceptReferenceTerms = new HashMap<>();
+
+	// Cache for item URL lookups - persists across clearCache() calls since items are used for metadata only
+	private final Map<String, Item> itemsByUrl = new HashMap<>();
+	private final Set<String> checkedItemUrls = new HashSet<>();
+	private boolean skipDbItemLookups = false;
 
 	/**
 	 * Pre-loads all concept sources into the cache to avoid repeated database queries.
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/openconceptlab/CacheService.java
```diff
@@ -59,6 +71,11 @@ public CacheService(ConceptService conceptService, OclConceptService oclConceptS
 	// Cache for ConceptReferenceTerms keyed by "sourceId:code"
 	Map<String, ConceptReferenceTerm> conceptReferenceTerms = new HashMap<>();
 
+	// Cache for item URL lookups - persists across clearCache() calls since items are used for metadata only
+	private final Map<String, Item> itemsByUrl = new HashMap<>();
+	private final Set<String> checkedItemUrls = new HashSet<>();
```
ibacher: I think this could use an explanation and, more precisely, why it wouldn't be ok to just use `itemsByUrl` with a `null` value (since that's pretty much how `HashSet` is implemented).

ibacher (commented):
I like this a lot! Please feed my frustration with AI comments to your coding agent.

ibacher:
Alright, so I've actually done something closer to real-world testing with this loading the PIH concept dictionary since that was the largest single dictionary I could find. With these changes (admittedly using the H2 database), this is about 1.7x faster (~6 minutes vs ~10 minutes). I've pushed my suggestions and some additional tests I wrote while trying to make some additional improvements (those attempts actually resulted in marginally slower load times, so I've abandoned them).

PR #123, opened 2026-04-02
OCLOMRS-1123: Add a configurable task to clean-up old import records

ibacher:
The goal here is to add a clean-up task that removes old imports in a somewhat configurable way. Basically, via GP, it's possible to specify a number of _days_ or a number of _runs_ that should be preserved. 

Regardless, the clean-up will preserve _at least one non-errored item for each URL_ so that the lookup via OCL URLs will continue to work.

This is a draft because it needs some testing and some clean-up.


ibacher:
Inadvertently closed while cleaning up some old worktrees.

PR #128, opened 2026-06-12
OCLOMRS-1123: Add configurable scheduled task to clean up old import records from the database

wikumChamith:
Add ImportCleanupTask, an OpenMRS scheduled task that purges import records older than a configurable retention period, including their openconceptlab_item rows


api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -579,4 +587,80 @@ public void flushAndClearSession() {
...
+		if (retentionDays <= 0) throw new IllegalArgumentException("retentionDays must be greater than 0, given: " + retentionDays);
+		if (batchSize <= 0) throw new IllegalArgumentException("batchSize must be greater than 0, given: " + batchSize);
+
+		ImportService importService = Context.getService(ImportService.class);
+		
+		Date cutoff = DateUtils.addDays(new Date(), -retentionDays);
+		Import lastSuccessful = importService.getLastSuccessfulSubscriptionImport();
+		Long excludedImportId = (lastSuccessful != null) ? lastSuccessful.getImportId() : null;
+
+		List<Import> purgable = importService.getImportsStoppedBefore(cutoff, excludedImportId, batchSize);
```
dkayiwa: issue (blocking): make `batchSize` a purge budget that pages through candidates — as written, retained imports permanently clog the oldest-N window and cleanup stops purging.

If merged as-is, cleanup permanently stops purging on long-lived instances, because imports that are kept forever (ones holding the last usable record for URLs no longer in the subscribed collection — the "Trimmed" case below) always occupy the front of this oldest-first, `LIMIT batchSize` window, and unpurgeable entries consume the entire batch on every run. This is the steady state the queue converges to: purgeable imports leave the front as they're purged, retained ones never do. Once `batchSize` of them have accumulated, every subsequent run fetches the same unpurgeable list and purges nothing — silently (the log just says "purged 0 import(s)"). An instance with old zip loads hits this even faster, since a zip import holding thousands of only-record URLs is retained from day one.

I verified this empirically on the branch (passes deterministically, both calls purge 0 and the fully superseded import is never reached):

```java
@Test
public void purgeOldImports_demoStarvation() throws Exception {
	createFailedImport(400, "/concepts/only-a/"); // only record for its URL -> retained forever
	createFailedImport(390, "/concepts/only-b/"); // only record for its URL -> retained forever
	Import superseded = createFailedImport(300, "/concepts/x/");
	createSuccessfulImport(1, "/concepts/x/");

	assertEquals(0, importService.purgeOldImports(90, 2));
	assertEquals(0, importService.purgeOldImports(90, 2)); // every later run stalls the same way
	assertNotNull(importService.getImport(superseded.getImportId())); // fully superseded, never examined
}
```

Suggested fix: keep fetching candidate pages (e.g. advance an offset by `batchSize`, or keyset-paginate on `localDateStopped`/`importId`) until `purged == batchSize` or no candidates remain. Trim-only imports still get their one-time trim on first encounter and are cheap to re-examine afterwards (two indexed queries each); if you want a hard bound on per-run work, cap the pages examined (e.g. 10×batchSize) — I'd still recommend the budget semantics as the default.

api/src/test/java/org/openmrs/module/openconceptlab/ImportCleanupServiceTest.java
```diff
@@ -0,0 +1,230 @@
...
+		Import recentImport = createFailedImport(5, "/concepts/a/");
+
+		int purged = importService.purgeOldImports(90, 10);
+
+		assertEquals(0, purged);
+		assertNotNull(importService.getImport(recentImport.getImportId()));
+	}
+
+	@Test
+	public void purgeOldImports_shouldNeverPurgeLastSuccessfulSubscriptionImport() throws Exception {
```
dkayiwa, line 78: suggestion (non-blocking): add a test that actually pins the `excludedImportId` guard — this one passes even with the guard deleted.

I mutation-tested it: with `excludedImportId` forced to `null` in `purgeOldImports`, all 14 new tests still pass, including this one. It survives because the last successful import here is also the *newest* import, so its item is the newest record for its URL and the `itemsRetained > 0` path keeps the import regardless — the exclusion never gets exercised. The guard is load-bearing (it's what protects `Importer.runTask()`'s incremental sync: if the last successful subscription import is purged, `getLastSuccessfulSubscriptionImport()` goes null and the next run silently falls back to a full re-import), so a future refactor could remove it with a green build.

The scenario that pins it needs a *newer failed* import to supersede all of the protected import's items. This version passes on your branch and fails when the guard is removed (verified both ways):

```java
@Test
public void purgeOldImports_shouldKeepLastSuccessfulSubscriptionImportEvenIfAllItsItemsAreSuperseded() throws Exception {
	Import oldSuccessfulImport = createSuccessfulImport(400, "/concepts/a/");
	createFailedImport(200, "/concepts/a/");

	int purged = importService.purgeOldImports(90, 10);

	assertEquals(0, purged);
	assertNotNull(importService.getImport(oldSuccessfulImport.getImportId()));
	assertEquals(1, itemCount(oldSuccessfulImport));
}
```

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -579,4 +587,80 @@ public void flushAndClearSession() {
...
+		if (excludedImportId != null) {
+			criteria.add(Restrictions.ne("importId", excludedImportId));
+		}
+		criteria.addOrder(Order.asc("localDateStopped"));
+		criteria.setMaxResults(maxResults);
+		return criteria.list();
+	}
+
+	@Override
+	public int purgeOldImports(int retentionDays, int batchSize) {
```
dkayiwa, line 637: suggestion (non-blocking): consider one transaction per purged import rather than one for the whole batch.

This method runs as a single `@Transactional` unit, so with the default batch of 10 and full-import-sized records (a CIEL-scale import carries a few hundred thousand items), one task run can delete a couple of million `openconceptlab_item` rows in a single MySQL transaction. The 1000-row chunking in `deleteItems()` bounds the IN-list, but not the transaction — locks are held and undo accumulates until the whole batch commits, and a failure near the end rolls back all completed work. A per-import transaction (e.g. loop in the task calling a purge-one-import service method, so each top-level service call gets its own transaction) bounds the worst case and preserves partial progress; it also composes naturally with the candidate paging suggested above. Given the conservative defaults this isn't blocking, but I'd recommend it before anyone cranks `importCleanupBatchSize` up to drain a large backlog.

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -579,4 +587,80 @@ public void flushAndClearSession() {
...
+		Date cutoff = DateUtils.addDays(new Date(), -retentionDays);
+		Import lastSuccessful = importService.getLastSuccessfulSubscriptionImport();
+		Long excludedImportId = (lastSuccessful != null) ? lastSuccessful.getImportId() : null;
+
+		List<Import> purgable = importService.getImportsStoppedBefore(cutoff, excludedImportId, batchSize);
+		int purged = 0;
+
+		for (Import anImport : purgable) {
+			int itemsDeleted = deleteItems(getDeletableItemIds(anImport));
+			int itemsRetained = importService. getImportItemsCount(anImport, Collections.emptySet());
```
dkayiwa: nit: stray space after the dot.

```suggestion
			int itemsRetained = importService.getImportItemsCount(anImport, Collections.emptySet());
```

api/src/main/java/org/openmrs/module/openconceptlab/scheduler/ImportCleanupTask.java
```diff
@@ -0,0 +1,64 @@
+package org.openmrs.module.openconceptlab.scheduler;
```
dkayiwa, line 10: nit: missing the MPL license header (every other file in the module has it), and this file uses 4-space indentation where the rest of the codebase uses tabs (`eclipse/OpenMRSFormatter.xml` via the formatter plugin). While at it: `ImportCleanupTaskSetup.java`, `ImportCleanupServiceTest.java` and `ImportCleanupTaskTest.java` are missing a trailing newline.

dkayiwa (commented):
Thanks @wikumChamith — this is a carefully designed cleanup feature, and most of it holds up very well under heavy scrutiny. I went beyond reading the diff and verified the behavior empirically; details below. There is **one blocking issue** (the batch window can permanently stall behind retained imports), one test-coverage gap on a load-bearing guard, and a couple of smaller points.

### What I verified empirically

**Builds & the surefire pin.** I built `master` and this branch in throwaway worktrees and compared test runs: 126 tests on master vs 140 on this branch — +14, exactly the new `ImportCleanupServiceTest` (11) + `ImportCleanupTaskTest` (3). Every pre-existing (JUnit 4) test class runs an identical count under surefire 2.22.2, because `openmrs-test:2.7.8` puts `junit-vintage-engine` on the classpath. The pin is also what makes the new JUnit 5 tests run at all on Mavens whose default surefire is 2.12.4 — without it they'd be silently skipped. So the pom change is safe and necessary. ✅

**Resolution semantics are preserved.** The purge HQL deletes an item only if it's `ERROR` or a newer non-error item exists with the same `hashedUrl` + `url` (higher `itemId`). That is exactly the complement of `getLastSuccessfulItemByUrl()` (filters `state != ERROR`, same `hashedUrl`+`url` match, picks max `itemId`) — so the one item that method can return is never deleted. Traced all consumers: `Saver`/`CacheService` URL resolution, `Importer.runTask()` incremental sync (`oclDateStarted`/`releaseVersion` are only ever read off `getLastSuccessfulSubscriptionImport()`, which the purge excludes), `ImportResource` (history only), and `HighlightSubscribedConcept.isSubscribedConcept()` (the surviving newest item preserves the uuid). ✅

**FK safety.** `openconceptlab_item.import_id` has an FK with no cascade; `getImportItemsCount(anImport, emptySet())` counts all states, so the `itemsRetained == 0` guard genuinely means zero rows remain before `session.delete()`. In-progress imports are excluded by SQL `<` semantics on NULL `localDateStopped` (and covered by a test). ✅

**Scheduler lifecycle.** Checked against openmrs-core 2.7.8 sources: `TimerSchedulerServiceImpl.scheduleTask()` shuts down an existing scheduled instance for the same task id, so `registerTask()` in `started()` plus core's own startup scheduling cannot double-schedule. First execution lands ~24h after registration (`SchedulerUtil.getNextExecution`), and re-starts honor an admin's `startOnStartup` change because the definition is only created when absent. ✅

**Mutation test on the exclusion guard.** I deleted the `excludedImportId` exclusion and re-ran the suite: **all 14 new tests still pass**, including `purgeOldImports_shouldNeverPurgeLastSuccessfulSubscriptionImport` — see the inline comment for why, plus a verified test that does pin it.

**Batch-window starvation.** I wrote a test demonstrating that permanently-retained imports clog the oldest-N candidate window and stop all purging — see the blocking inline comment.

### Needs action before merge

- [ ] Make `batchSize` a purge budget (page through candidates) so permanently-retained imports can't permanently stall the cleanup (inline comment on `purgeOldImports`).

Everything else is non-blocking. Nice work on the select-then-delete workaround for MySQL error 1093 and on documenting it, and the test helpers made my verification work much easier.

dkayiwa:
@wikumChamith i am trying to come up with a skill that responds to your AI review concerns starting from the previous pull request in the events module. Therefore, feel free to give me feedback about what you do not like about the review for this pull request and i refine the skill even further. My goal is to come up with a final version of the skill which does better than my personal review, or at least close. Of course i will always do a final manual review, but i just want to leverage AI as much as possible, but in a meaningful way that is relevant to the OpenMRS project.

wikumChamith:
@dkayiwa I updated the PR :)

dkayiwa:
@wikumChamith did you get a chance to look at the sonar issues? :)

wikumChamith:
> @wikumChamith did you get a chance to look at the sonar issues? :)

I actually fixed them before. Not sure why Sonar bot hasn't put a new comment :)

wikumChamith:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-openconceptlab&pullRequest=128) **Quality Gate passed**
> 
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=openmrs_openmrs-module-openconceptlab&pullRequest=128&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=openmrs_openmrs-module-openconceptlab&pullRequest=128&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=openmrs_openmrs-module-openconceptlab&pullRequest=128&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  No data about Coverage  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=openmrs_openmrs-module-openconceptlab&pullRequest=128&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-openconceptlab&pullRequest=128)

This remaining issue is a false positive.

dkayiwa (commented):
Disposition: **merge** (re-review against 5daca47) — nothing blocking remains.

Two non-blocking leftovers, neither gating merge:
- No test exercises `tryPurgeImport`'s catch-and-continue path — the behaviour that keeps one failed purge from aborting the whole batch (inline comment).
- `ImportCleanupTask.java` still lacks a trailing newline (noted in the existing formatting thread).

Needs action before merge: none.

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -579,4 +600,116 @@ public void flushAndClearSession() {
...
+	 * @return {@code true} if the import was purged, {@code false} if it was kept or failed to purge
+	 */
+	private boolean tryPurgeImport(ImportService importService, Long importId) {
+		try {
+			return importService.purgeImport(importId);
+		}
+		catch (Exception e) {
+			log.error("Failed to purge OCL import id={}, skipping it and continuing", importId, e);
+			return false;
+		}
```
dkayiwa, line 682: suggestion (non-blocking): add a test that pins `tryPurgeImport`'s catch-and-continue, so a future refactor can't silently turn one failed purge into a whole-batch abort.

Nothing in the 15 service tests exercises this catch — `purgeImport` only ever succeeds in them, so the catch block (and the `kept++` that lets the batch skip a failed import and keep going) is dead in the suite. It's the same silent-regression shape as the `excludedImportId` guard from last round: the per-import-transaction split's stated point is "earlier purges survive if a later one fails," and that guarantee currently passes CI whether this try/catch is here or not. A spy `ImportService` that throws from `purgeImport` for one id — asserting the batch still purges the others and returns the expected count — would lock it in. The code is correct as written; this is purely about pinning the behaviour.

api/src/main/java/org/openmrs/module/openconceptlab/scheduler/ImportCleanupTask.java
dkayiwa: nit: `ImportCleanupTask.java` still has no trailing newline at EOF — add one to match the other files.

wikumChamith:
@dkayiwa It would be great if we could avoid a comment if something is fixed and there are no problems in the solution :) I think that will help reduce the slop and make it feel more human-like.

dkayiwa:
@wikumChamith very useful feedback. Let me ask claude to directly read your advice and edit the skill accordingly. 👍

dkayiwa:
@wikumChamith claude has updated the skill and used it to edit the previous comments. Do they look right? Or should i refine it further?

wikumChamith:
> Disposition: **merge** (re-review against [5daca47](https://github.com/openmrs/openmrs-module-openconceptlab/commit/5daca4779ad7c7578a3140ae79f3b6d0df19b2b7)) — nothing blocking remains.
> 
> Two non-blocking leftovers, neither gating merge:
> 
>     * No test exercises `tryPurgeImport`'s catch-and-continue path — the behaviour that keeps one failed purge from aborting the whole batch (inline comment).
> 
>     * `ImportCleanupTask.java` still lacks a trailing newline (noted in the existing formatting thread).
> 
> 
> Needs action before merge: none.

@dkayiwa is this ready to get merged now :)

api/src/main/java/org/openmrs/module/openconceptlab/ImportServiceImpl.java
```diff
@@ -579,4 +600,116 @@ public void flushAndClearSession() {
...
+	@Override
+	public int purgeOldImports(int retentionDays, int batchSize) {
+		if (retentionDays <= 0) throw new IllegalArgumentException("retentionDays must be greater than 0, given: " + retentionDays);
+		if (batchSize <= 0) throw new IllegalArgumentException("batchSize must be greater than 0, given: " + batchSize);
+
+		ImportService importService = Context.getService(ImportService.class);
+
+		Date cutoff = Date.from(clock.instant().minus(retentionDays, ChronoUnit.DAYS));
+		Import lastSuccessful = importService.getLastSuccessfulSubscriptionImport();
+		Long excludedImportId = (lastSuccessful != null) ? lastSuccessful.getImportId() : null;
```
dkayiwa, line 645: While an import is running this resolves to the running import rather than to the last *completed* successful one, so the import `Importer.runTask()` would actually fall back to is left unprotected.

`getLastSuccessfulSubscriptionImport()` filters only on `errorMessage IS NULL` and `oclDateStarted IS NOT NULL`, and `runTask()` sets `oclDateStarted` before `processInput()`, so for the whole content-processing phase of an import the in-flight row is the highest-id match. `excludedImportId` then points at a row that the `localDateStopped < cutoff` predicate already excludes, and the previous successful import drops back into the candidate set.

I ran it on the branch with a control, so the difference is attributable to the running import. Both of these pass as written:

```java
@Test
public void control_guardProtectsLastSuccessfulWhenNothingIsRunning() throws Exception {
	Import oldSuccessful = createSuccessfulImport(400, "/concepts/a/");
	createFailedImport(200, "/concepts/a/"); // newer, supersedes its item

	assertEquals(0, importService.purgeOldImports(90, 10));
	assertNotNull(importService.getImport(oldSuccessful.getImportId()));
}

@Test
public void probe_runningImportShadowsGuard() throws Exception {
	Import oldSuccessful = createSuccessfulImport(400, "/concepts/a/");

	Import running = new Import(); // what runAndHandleErrors + runTask do, up to processInput()
	importService.startImport(running);
	importService.updateOclDateStarted(running, Date.from(NOW));
	addItem(running, "/concepts/a/", ItemState.ADDED);

	assertEquals(1, importService.purgeOldImports(90, 10));
	assertImportPurged(oldSuccessful.getImportId());
}
```

In the probe the guard resolves to the running import, the old one is purged, and once I then call `failImport(running, ...)` the next `getLastSuccessfulSubscriptionImport()` returns null, which is the silent fall back to a full re-import in `runTask()`. Nothing logs that it happened.

It needs the last successful import to be both older than the retention window and fully superseded by the time cleanup fires, so it is narrow (an instance that imports rarely, or one whose subscription has been failing for months and is now retrying), and the cost is a slow re-import rather than lost data. That is why I would not hold the merge for it.

What I would change: resolve the exclusion to the last successful subscription import that has actually stopped, i.e. the same criteria plus `localDateStopped IS NOT NULL`. Because `purgeOldImports` deliberately runs outside a transaction, that has to be a service-level lookup rather than a `getSession()` call here. Skipping the whole run while `getInProgressImports()` is non-empty would also close it and is less code, but it lets a long import starve the cleanup, so I would go with the query change.

dkayiwa (commented):
@wikumChamith yes, nothing here blocks it. The one comment I left on `purgeOldImports` is worth fixing but is not a merge gate.

PR #129, opened 2026-06-16
OCLOMRS-1126: Propagate `retired` and `retire_reason` from OCL when importing concepts and mappings

jwnasambu:
**Description of what I fixed**
The fix on this PR ensures that when importing concepts and mappings from OCL their retirement status and reasons are correctly synchronized with OpenMRS by reading the OCL retired and retire_reason fields. Specifically, I updated the Saver class to apply these fields to concepts, concept names and reference-term mappings, while retired concept descriptions  correctly removed to align with OpenMRS semantics. The implementation supports reactivation by unretiring/unvoiding objects and clearing reasons when OCL's status changes to active, and it maintains backwards compatibility by defaulting missing reason fields to "Retired in OCL."

**Issue worked on:**
https://openmrs.atlassian.net/browse/OCLOMRS-1126

### DIFF
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
@@ -45,6 +45,9 @@ public class OclConcept {
 
 	private boolean retired;
 
+	@JsonProperty("retire_reason")
+	private String retireReason;
+
 	private String url;
 
 	@JsonProperty("version_url")
@@ -133,6 +136,14 @@ public class OclConcept {
 
 	public void setRetired(boolean retired) {
 		this.retired = retired;
+	}
+
+	public String getRetireReason() {
+		return retireReason;
+	}
+
+	public void setRetireReason(String retireReason) {
+		this.retireReason = retireReason;
 	}
 
 	public String getUrl() {
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
@@ -222,6 +222,11 @@ public class OclConcept {
 
 		private String type;
 
+		private Boolean retired;
+
+		@JsonProperty("retire_reason")
+		private String retireReason;
+
 		public String getUuid() {
 			return uuid;
 		}
@@ -278,6 +283,26 @@ public class OclConcept {
 			this.type = type;
 		}
 
+		public Boolean getRetired() {
+			return retired;
+		}
+
+		public boolean isRetired() {
+			return Boolean.TRUE.equals(retired);
+		}
+
+		public void setRetired(Boolean retired) {
+			this.retired = retired;
+		}
+
+		public String getRetireReason() {
+			return retireReason;
+		}
+
+		public void setRetireReason(String retireReason) {
+			this.retireReason = retireReason;
+		}
+
 		@Override
 		public String toString() {
 			return new ToStringBuilder(this).append("name", name).build();
@@ -324,6 +349,11 @@ public class OclConcept {
 
 		private String description;
 
+		private Boolean retired;
+
+		@JsonProperty("retire_reason")
+		private String retireReason;
+
 		public String getUuid() {
 			return uuid;
 		}
@@ -354,6 +384,26 @@ public class OclConcept {
 
 		public void setDescription(String description) {
 			this.description = description;
+		}
+
+		public Boolean getRetired() {
+			return retired;
+		}
+
+		public boolean isRetired() {
+			return Boolean.TRUE.equals(retired);
+		}
+
+		public void setRetired(Boolean retired) {
+			this.retired = retired;
+		}
+
+		public String getRetireReason() {
+			return retireReason;
+		}
+
+		public void setRetireReason(String retireReason) {
+			this.retireReason = retireReason;
 		}
 
 		public void copyFrom(ConceptDescription description) {
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
@@ -203,14 +203,7 @@ public class OclConcept {
 	}
 
 	@JsonIgnoreProperties(ignoreUnknown = true)
-	public static class Name {
-
-		private String uuid;
-
-		@JsonProperty("external_id")
-		private String externalId;
-
-		private Locale locale;
+	public static class Name extends OclLocalizedObject {
 
 		@JsonProperty("locale_preferred")
 		private boolean localePreferred;
@@ -221,35 +214,6 @@ public class OclConcept {
 		private String nameType;
 
 		private String type;
-
-		private Boolean retired;
-
-		@JsonProperty("retire_reason")
-		private String retireReason;
-
-		public String getUuid() {
-			return uuid;
-		}
-
-		public void setUuid(String uuid) {
-			this.uuid = uuid;
-		}
-
-		public String getExternalId() {
-			return externalId;
-		}
-
-		public void setExternalId(String externalId) {
-			this.externalId = externalId;
-		}
-
-		public Locale getLocale() {
-			return locale;
-		}
-
-		public void setLocale(Locale locale) {
-			this.locale = locale;
-		}
 
 		public boolean isLocalePreferred() {
 			return localePreferred;
@@ -283,26 +247,6 @@ public class OclConcept {
 			this.type = type;
 		}
 
-		public Boolean getRetired() {
-			return retired;
-		}
-
-		public boolean isRetired() {
-			return Boolean.TRUE.equals(retired);
-		}
-
-		public void setRetired(Boolean retired) {
-			this.retired = retired;
-		}
-
-		public String getRetireReason() {
-			return retireReason;
-		}
-
-		public void setRetireReason(String retireReason) {
-			this.retireReason = retireReason;
-		}
-
 		@Override
 		public String toString() {
 			return new ToStringBuilder(this).append("name", name).build();
@@ -310,7 +254,7 @@ public class OclConcept {
 
 		@Override
 		public int hashCode() {
-			return new HashCodeBuilder().append(locale).append(localePreferred).append(name).append(nameType).build();
+			return new HashCodeBuilder().append(getLocale()).append(localePreferred).append(name).append(nameType).build();
 		}
 
 		@Override
@@ -325,58 +269,22 @@ public class OclConcept {
 				return false;
 			}
 			Name rhs = (Name) obj;
-			return new EqualsBuilder().append(locale, rhs.locale).append(localePreferred, rhs.localePreferred)
+			return new EqualsBuilder().append(getLocale(), rhs.getLocale()).append(localePreferred, rhs.localePreferred)
 			        .append(name, rhs.name).append(nameType, rhs.nameType).build();
 		}
 
 		public void copyFrom(ConceptName name) {
 			this.name = name.getName();
-			locale = name.getLocale();
+			setLocale(name.getLocale());
 			localePreferred = name.getLocalePreferred() != null ? name.getLocalePreferred() : false;
 			nameType = name.getConceptNameType() != null ? name.getConceptNameType().toString() : null;
 		}
 	}
 
 	@JsonIgnoreProperties(ignoreUnknown = true)
-	public static class Description {
-
-		private String uuid;
-
-		@JsonProperty("external_id")
-		private String externalId;
-
-		private Locale locale;
+	public static class Description extends OclLocalizedObject {
 
 		private String description;
-
-		private Boolean retired;
-
-		@JsonProperty("retire_reason")
-		private String retireReason;
-
-		public String getUuid() {
-			return uuid;
-		}
-
-		public void setUuid(String uuid) {
-			this.uuid = uuid;
-		}
-
-		public String getExternalId() {
-			return externalId;
-		}
-
-		public void setExternalId(String externalId) {
-			this.externalId = externalId;
-		}
-
-		public Locale getLocale() {
-			return locale;
-		}
-
-		public void setLocale(Locale locale) {
-			this.locale = locale;
-		}
 
 		public String getDescription() {
 			return description;
@@ -386,34 +294,14 @@ public class OclConcept {
 			this.description = description;
 		}
 
-		public Boolean getRetired() {
-			return retired;
-		}
-
-		public boolean isRetired() {
-			return Boolean.TRUE.equals(retired);
-		}
-
-		public void setRetired(Boolean retired) {
-			this.retired = retired;
-		}
-
-		public String getRetireReason() {
-			return retireReason;
-		}
-
-		public void setRetireReason(String retireReason) {
-			this.retireReason = retireReason;
-		}
-
 		public void copyFrom(ConceptDescription description) {
 			this.description = description.getDescription();
-			locale = description.getLocale();
+			setLocale(description.getLocale());
 		}
 
 		@Override
 		public int hashCode() {
-			return new HashCodeBuilder().append(locale).append(description).build();
+			return new HashCodeBuilder().append(getLocale()).append(description).build();
 		}
 
 		@Override
@@ -428,12 +316,12 @@ public class OclConcept {
 				return false;
 			}
 			Description rhs = (Description) obj;
-			return new EqualsBuilder().append(locale, rhs.locale).append(description, rhs.description).build();
+			return new EqualsBuilder().append(getLocale(), rhs.getLocale()).append(description, rhs.description).build();
 		}
 
 		@Override
 		public String toString() {
-			return new ToStringBuilder(this).append("description", description).append("locale", locale).build();
+			return new ToStringBuilder(this).append("description", description).append("locale", getLocale()).build();
 		}
 	}
 
--- a/api/src/main/java/org/openmrs/module/openconceptlab/client/OclMapping.java
+++ b/api/src/main/java/org/openmrs/module/openconceptlab/client/OclMapping.java
@@ -108,6 +108,7 @@ public class OclMapping {
 	public void setRetired(Boolean retired) {
 		this.retired = retired;
 	}
+
 	
 	public String getMapType() {
 		return mapType;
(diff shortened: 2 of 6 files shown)


solomonfortune:
Hey @jwnasambu , SonarQube Cloud has suggested some 7 issues regarding your PR. Is it possible for you to review and see what the concerns are?

jwnasambu:
@solomonfortune Thanks for the point out I am currently reviewing the SonarQube findings to understand the reported issues. I'll make the necessary fixes where appropriate and provide an update on the PR shortly.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -441,6 +454,17 @@ public Item saveMapping(final CacheService cacheService, final Import update, fi
...
+								// If it's not retired in OpenMRS but retired in OCL, retire it
+								if (BooleanUtils.isTrue(oclMapping.getRetired())) {
+									term.setRetired(true);
+									if (StringUtils.isNotBlank(oclMapping.getRetireReason())) {
+										term.setRetireReason(oclMapping.getRetireReason());
+									} else {
+										term.setRetireReason(OpenConceptLabConstants.DEFAULT_RETIRE_REASON);
+									}
+									conceptService.saveConceptReferenceTerm(term);
+								}
```
ibacher: While retiring the mapping is valid, I don't think we want to retire the concept reference term. This is because CRTs can be shared for multiple mappings (e.g., where two concepts map to the same term in another source, we try to use the same CRT). Consequently, retiring it without confirming this term is not otherwise in use is invalid.

ibacher (changes requested):
Thanks @jwnasambu! One adjustment we need to make and I think this will be good to go.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
jwnasambu: Just seen this. Let me fix it right away

jwnasambu: @ibacher Apologies for the delay in addressing the review. I had not seen the changes requested earlier but I have now implemented the requested changes.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -429,6 +432,14 @@ public Item saveMapping(final CacheService cacheService, final Import update, fi
...
 							term.setConceptSource(toSource);
 							term.setCode(oclMapping.getToConceptCode());
+							if (BooleanUtils.isTrue(oclMapping.getRetired())) {
+								term.setRetired(true);
+								if (StringUtils.isNotBlank(oclMapping.getRetireReason())) {
+									term.setRetireReason(oclMapping.getRetireReason());
+								} else {
+									term.setRetireReason(OpenConceptLabConstants.DEFAULT_RETIRE_REASON);
+								}
+							}
```
ibacher: Thanks @jwnasambu, but this block also repeats the same issue.

jwnasambu: Let me look into it right away!

jwnasambu:
@chibongho , @ibacher , @wikumChamith , @dkayiwa kindly help me review this PR at your convenient time please.

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -1,1803 +1,1856 @@
...
-		assertThat(importedConceptWithIndexTerm.getNames(), hasItem((Matcher<? super ConceptName>) allOf(hasProperty("conceptNameType", equalTo(ConceptNameType.INDEX_TERM)),
-			hasProperty("name", equalTo("Nazwa")))));
-	}
-	
-	/**
-	 * @see Saver#saveConcept(CacheService, Import, OclConcept)
-	 * @verifies save concept if versionUrl changed from last anImport
-	 */
-	@Test
-	public void importConcept_shouldUpdateConceptIfVersionUrlChanged() throws Exception {
```
dkayiwa, line 814: Before merge, I think these three tests need to come back. The file was rewritten here (it also flipped from CRLF to LF line endings, which is why it shows as 3,600+ changed lines), and in the process three tests that have nothing to do with retirement got dropped:

- `importConcept_shouldUpdateConceptIfVersionUrlChanged`
- `importConcept_shouldSkipUpdatingConceptIfVersionUrlDidntChange`
- `importConcept_shouldChangeDuplicateFullySpecifiedNamesToIndexTerm`

Losing them removes the only regression coverage for the version-URL update/skip optimization and for duplicate fully-specified-name handling, so if either of those paths breaks later nothing will catch it. It looks like accidental collateral of the line-ending rewrite rather than an intentional removal.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclMapping.java
```diff
@@ -28,6 +28,9 @@ public class OclMapping {
 	private String url;
 	
 	private Boolean retired;
+
+	@JsonProperty("retire_reason")
+	private String retireReason;
```
dkayiwa: This `retireReason` is parsed from OCL but never read anywhere (`OclMapping` is only ever deserialized, and nothing calls `getRetireReason()`). A retired mapping is handled in `Saver.saveMapping` by removing the `ConceptMap`, and a `ConceptMap` has no retire-reason to store, so unlike the concept and name reasons this one has nowhere to land. Since the PR title covers mappings, it's easy to read this as wired up when it isn't. I'd drop the field and its getter/setter unless you have a use planned for it.

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -1,1803 +1,1856 @@
...
+				break;
+			}
+		}
+		assertThat(unvoidedName, is(notNullValue()));
+		assertFalse(unvoidedName.isVoided());
+		assertNull(unvoidedName.getVoidReason());
+	}
+
+	@Test
+	public void saveMapping_shouldUnretireReferenceTermAndClearReason() throws Exception {
```
dkayiwa: This doesn't actually exercise the unretire path. `saveMapping` never retires a concept reference term (that's the behavior from the earlier review round), so the term created in the first call stays unretired, and the `if (BooleanUtils.isTrue(term.getRetired()))` branch in `Saver` is never entered on the second call. Both assertions then hold trivially, whether or not that branch works. To make it meaningful, retire the term first (e.g. `conceptService.retireConceptReferenceTerm(term, "...")` on the imported term), then import the non-retired mapping and assert it comes back unretired with a null reason.

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -1,1803 +1,1856 @@
...
+		oclConcept.getDescriptions().get(0).setRetired(true);
+
+		saver.saveConcept(new CacheService(conceptService, oclConceptService), update, oclConcept);
+
+		Concept concept = conceptService.getConceptByUuid(oclConcept.getExternalId());
+		assertThat(concept.getDescriptions(), is(empty()));
+	}
+
+	@Test
+	public void saveMapping_shouldRetireReferenceTermWithReason() throws Exception {
```
dkayiwa: The name and the assertion here disagree: the method is `saveMapping_shouldRetireReferenceTermWithReason`, but it asserts `assertFalse(term.getRetired())`, i.e. that the term is not retired. The assertion is the right contract (a reference term can be shared across mappings, so it shouldn't be retired just because one mapping is); it's the name that says the opposite. Renaming to something like `shouldNotRetireReferenceTermWhenMappingIsRetired` would make it match what it checks. Not important, just a bit confusing to read.

jwnasambu:
@dkayiwa Thanks for the review I have fixed the proposed chnages.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
```diff
@@ -211,6 +222,11 @@ public static class Name {
 
 		private String type;
 
+		private boolean retired;
```
dkayiwa: Does OCL actually send `retired` / `retire_reason` on individual names and descriptions? I checked the three OCL export fixtures we test against (`response.zip` plus both `refapp-collection-response` files): the concept level has `retired` on all 100 concepts, but none of the 305 names or 101 descriptions carry `retired` or `retire_reason` at all. If that reflects the real OCL format, then `oclName.isRetired()` and `oclDescription.isRetired()` are always false on a real import, and the name-voiding here plus the retired-description handling in `Saver` never run outside the unit tests that set the flags by hand.

The module also already voids names and removes descriptions that OCL drops from the concept's list (`voidNamesRemovedFromOcl` / `removeDescriptionsRemovedFromOcl`), which is the usual way OCL signals a name or description is gone, so that scenario looks covered regardless.

Could you confirm against a live export that these per-name/per-description fields exist? If they don't, I'd drop the `retired`/`retire_reason` fields on `Name`/`Description` and their `Saver` branches and keep just the concept-level reason change. These fixtures are from 2015-2016, so I can't rule out a newer OCL version adding the fields, which is why I'm asking rather than calling it dead.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -639,10 +641,20 @@ private void updateOrAddNamesFromOcl(Concept concept, OclConcept oclConcept) {
 					name.setConceptNameType(oclNameType);
 					name.setLocalePreferred(oclName.isLocalePreferred());
 
-					//Unvoiding if necessary
-					name.setVoided(false);
-					name.setVoidReason(null);
-					name.setVoidedBy(null);
+					// Handle voiding if retired in OCL
+					if (oclName.isRetired()) {
```
dkayiwa, line 652: If those `retired` fields do end up being populated (see my note on `OclConcept.Name`), it's worth knowing this branch has no test, though I wouldn't block on it. The case is an existing name that OCL newly flags as retired: `saveConcept_shouldRetireNameWithReason` covers the new-name branch just below (retired on first import) and `saveConcept_shouldUnvoidNameAndClearReason` covers unvoiding, but nothing imports an active name and then re-imports it retired, which is the normal re-sync path. The `!oclDescription.isRetired()` clause in `removeDescriptionsRemovedFromOcl` (line 776) has the same gap. I reverted both branches to their master behavior and the full SaverTest still passed 62/62, so either could be dropped in a refactor without a test noticing. A two-import test for each (import active, re-import retired, assert voided / removed) would cover it.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -639,11 +641,6 @@ private void updateOrAddNamesFromOcl(Concept concept, OclConcept oclConcept) {
 					name.setConceptNameType(oclNameType);
 					name.setLocalePreferred(oclName.isLocalePreferred());
 
-					//Unvoiding if necessary
```
wikumChamith, line 642: Why are we removing this?

jwnasambu:
@dkayiwa, @ibacher kindly help me review this PR at your convenient time please!

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -231,7 +231,11 @@ public Concept toConcept(CacheService cacheService, OclConcept oclConcept) throw
 
 		concept.setRetired(oclConcept.isRetired());
 		if (oclConcept.isRetired()) {
-			concept.setRetireReason("Retired in OCL");
+			if (StringUtils.isNotBlank(oclConcept.getRetireReason())) {
+				concept.setRetireReason(oclConcept.getRetireReason());
+			} else {
+				concept.setRetireReason(OpenConceptLabConstants.DEFAULT_RETIRE_REASON);
+			}
```
dkayiwa: OCL's `retire_reason` is unbounded upstream (a Django `TextField`), while `concept.retire_reason` in OpenMRS is varchar(255). A longer reason will fail the concept save with a data-too-long error, and that concept then errors out of the import. Since the value comes from an external server I'd cap it:

```suggestion
			if (StringUtils.isNotBlank(oclConcept.getRetireReason())) {
				concept.setRetireReason(StringUtils.abbreviate(oclConcept.getRetireReason(), 255));
			} else {
				concept.setRetireReason(OpenConceptLabConstants.DEFAULT_RETIRE_REASON);
			}
```

Take it or leave it, this doesn't block anything for me.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
dkayiwa: It turns out these per-name/description fields are real in current OCL, so the premise of my question above was stale. OCL added `retire_reason` to concepts, mappings, and concept names/descriptions in June (OpenConceptLab/ocl_issues#2519, oclapi2 commit b26ea69d), and the export serializers (`ConceptVersionExportSerializer`, which nests `LocalizedNameSerializer`/`LocalizedDescriptionSerializer`) now emit `retired` and `retire_reason` on every name and description. The production API is already on a build 53 commits past that change; our test fixtures were just showing their age. Sorry for the bad steer.

This matters for scope, because OCLOMRS-1126's acceptance criteria explicitly include per-name/description retirement, and this revision dropped that handling in response to my comment. As the code now stands, a name that OCL retires still imports as active: it stays in the export's names list, so `updateOrAddNamesFromOcl` matches it and unconditionally unvoids it, and the removed-from-list voiding never applies. If this merges as-is and the ticket closes, subscribed dictionaries keep showing names OCL has retired (the exact gap the ticket describes), and the PR title and description would claim name/description propagation the code no longer does.

I'd bring the name/description handling back from the earlier revision (it still exists on commit 3549c0f1fdec9a0ba5776c23abaa2ad765f3084c, reachable from this PR's force-push timeline entry), together with the name/description tests the acceptance criteria list. If maintainers would rather land concept-level now and split names/descriptions into a follow-up, that works too, but then the ticket and the PR title/description need to say so before this merges. My preference is restoring it here, since the code was already written once.

The mapping half needs nothing more in my view: retired-state is already reflected by the OCLOMRS-1107 code (the ConceptMap is removed on retire, re-added and the term unretired on reactivation), and mapping `retire_reason` has nowhere to land because ConceptMap has no reason field.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclMapping.java
```diff
@@ -108,6 +108,7 @@ public boolean isRetired() {
 	public void setRetired(Boolean retired) {
 		this.retired = retired;
 	}
+
```
ibacher: nit: Can we remove this change? Doesn't seem helpful.

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -376,7 +380,6 @@ public Item saveMapping(final CacheService cacheService, final Import update, fi
 						if (toItem != null) {
 							toConcept = cacheService.getConceptByUuid(toItem.getUuid());
 						}
-
```
ibacher, line 379: This line, on the other hand shouldn't be cleared in this PR

api/src/main/java/org/openmrs/module/openconceptlab/importer/Saver.java
```diff
@@ -615,7 +618,6 @@ private Double getSortWeightForMapping(OclMapping oclMapping, Double defaultIfUn
 		if (sortWeight != null) {
 			return sortWeight;
 		}
-
```
ibacher, line 618: Ditto about not clearing this.

jwnasambu:
Thanks @dkayiwa and @dkayiwa for the review. I have fixed the proposed changes feel free to review my PR  at your convenient time please!

api/src/test/java/org/openmrs/module/openconceptlab/importer/SaverTest.java
```diff
@@ -1,1803 +1,1884 @@
...
+		Import update = importService.getLastImport();
+		OclConcept oclConcept = newOclConcept();
+		oclConcept.setRetired(true);
+		oclConcept.setRetireReason("Reason for retirement");
+
+		saver.saveConcept(new CacheService(conceptService, oclConceptService), update, oclConcept);
+
+		Concept concept = conceptService.getConceptByUuid(oclConcept.getExternalId());
+		assertTrue(concept.isRetired());
+		assertThat(concept.getRetireReason(), is("Reason for retirement"));
```
dkayiwa: Worth adding a companion test for the truncation branch: import a concept whose retire_reason is longer than 255 characters and assert the stored reason comes back as 255 characters ending in "...". Nothing currently fails if the StringUtils.abbreviate call gets dropped in a refactor, since the reason used here fits under the limit. Not something I would hold the PR on.

api/src/main/java/org/openmrs/module/openconceptlab/client/OclConcept.java
dkayiwa: This revision still doesn't have the name/description half: OclConcept.Name and OclConcept.Description carry no retired/retire_reason fields and Saver has no handling for them, while the PR title/description and the ticket still promise it (I re-checked OCLOMRS-1126 just now and the acceptance criteria haven't been re-scoped). So a name retired in OCL still imports as active, because updateOrAddNamesFromOcl unvoids any name that is still present in the export's list. My preference is still to bring that handling back from 3549c0f1fdec9a0ba5776c23abaa2ad765f3084c together with its tests. If maintainers would rather land concept-level only, then the ticket needs splitting and the title/description trimming to concept level before this merges, so OCLOMRS-1126 doesn't get closed with its main gap still open.

jwnasambu:
@dkayiwa, @ibacher Thanks for review. Kindly could there something else to be fixed?

ibacher (commented):
So the ticket requested handling `retired` and `retire_reason` in three places:

* Concepts
* Concept Names and Descriptions
* Concept Mappings

This PR only addresses the first of those.

jwnasambu:
@ibacher, @dkayiwa, @wikumChamith  kindly feel free to review my changes at your convenient time please

ibacher (approved):
Thanks @jwnasambu!

